PR #1, opened 2019-04-16
Initial port from lifecycle

ameyer-pivotal:
Left factory in lifecycle, added new constructors here:

- `EmptyLocalImage(...)`
- `NewLocalImage(...)`
- `NewRemoteImage(...)`

Signed-off-by: Andrew Meyer <ameyer@pivotal.io>


.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,12 @@
+<component name="ProjectCodeStyleConfiguration">
```
sclevine: Could we `.gitignore` the `.idea` directory? Would prefer not to include individual IDE prefs in the project.

go.mod
```diff
@@ -0,0 +1,27 @@
+module github.com/buildpack/imgutil
```
sclevine: Given that this is a library, I don't see a strong motivation for vendoring all the dependencies. Could we rip those out?

.idea/codeStyles/Project.xml
ameyer-pivotal: Good catch! Forgot to check that.

go.mod
ameyer-pivotal: Sure, no problem

.idea/codeStyles/Project.xml
ameyer-pivotal: Done.

go.mod
ameyer-pivotal: Done.

testhelpers/testhelpers.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func ImageID(t *testing.T, repoName string) string {
+	t.Helper()
+	inspect, _, err := DockerCli(t).ImageInspectWithRaw(context.Background(), repoName)
+	AssertNil(t, err)
+	return inspect.ID
+}
+
+// func Run(t *testing.T, cmd *exec.Cmd) string {
```
jchesterpivotal: Delete commented code?

ameyer-pivotal: @jchesterpivotal good catch ;) Done.

fakes/image.go
```diff
@@ -0,0 +1,260 @@
...
+	return &Image{
+		t:            t,
+		alreadySaved: false,
+		labels:       make(map[string]string),
+		env:          make(map[string]string),
+		topLayerSha:  topLayerSha,
+		digest:       digest,
+		name:         name,
+		cmd:          []string{"initialCMD"},
+		layersMap:    map[string]string{},
```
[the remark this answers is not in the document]
ameyer-pivotal: Done.

fakes/image.go
```diff
@@ -0,0 +1,260 @@
...
+		createdAt:    time.Now(),
```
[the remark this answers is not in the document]
ameyer-pivotal: I'm not sure about that one. If it becomes a problem, we can fix it later (this is just a copy from lifecycle, so it's been like that for a while).

remote.go
```diff
@@ -0,0 +1,328 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/pkg/errors"
+)
+
+type RemoteImage struct {
+	keychain   authn.Keychain
+	RepoName   string
```
ekcasey: I know these were already exported before we moved this over to this new repo, but is there any reason to expose RepoName, Image and PrevLayers? They seem like implementation details that should be encapsulated.

local.go
```diff
@@ -0,0 +1,549 @@
...
+	layerPaths       []string
+	Stdout           io.Writer
+	currentTempImage string
+	prevDir          string
+	prevMap          map[string]string
+	prevOnce         *sync.Once
+	easyAddLayers    []string
+}
+
+func EmptyLocalImage(repoName string, dockerClient *client.Client) *LocalImage {
```
ekcasey: Do we need to inject the docker client? Can we just instantiate it in the constructor?

local.go
```diff
@@ -0,0 +1,549 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/pkg/errors"
+)
+
+type LocalImage struct {
+	RepoName         string
+	Docker           *client.Client
+	Inspect          types.ImageInspect
+	layerPaths       []string
+	Stdout           io.Writer
```
ekcasey: It looks like Stdout isn't used, now that logging has been removed from the image

local.go
```diff
@@ -0,0 +1,549 @@
...
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/pkg/errors"
+)
+
+type LocalImage struct {
+	RepoName         string
+	Docker           *client.Client
+	Inspect          types.ImageInspect
```
ekcasey: Do these fields need to be exported?

local.go
sclevine: This actually seems nice to me. What if you want to instantiate the client with different parameters (e.g., with different daemon location)? Given that this is a library, probably better to avoid reading Docker env vars directly.

ekcasey: yeah that makes sense

remote.go
ameyer-pivotal: Done.

local.go
ameyer-pivotal: Done.

local.go
ameyer-pivotal: Done (unexported).

PR #10, opened 2019-07-25
Retrieve remote layers uncompressed and implement remote delete.

lukasberger:
The lifecycle image cache expects layers to be retrieved as uncompressed
tars and attempts to delete the original cache image when the new image
is written.

### DIFF
--- a/remote/remote_test.go
+++ b/remote/remote_test.go
@@ -737,6 +737,38 @@ func testRemoteImage(t *testing.T, when spec.G, it spec.S) {
 			})
 		})
 	})
+
+	when("#Delete", func() {
+		when("it exists", func() {
+			var img imgutil.Image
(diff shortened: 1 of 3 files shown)


remote/remote_test.go
```diff
@@ -737,6 +737,38 @@ func testRemoteImage(t *testing.T, when spec.G, it spec.S) {
...
+
+				var err error
+				img, err = remote.NewImage(
+					repoName, authn.DefaultKeychain,
+					remote.FromBaseImage(repoName),
+				)
+				h.AssertNil(t, err)
+			})
+
+			it("returns nil", func() {
```
ekcasey: I think this test should check that the image has actually been deleted.

lukasberger: Good point, added.

PR #14, opened 2019-10-14
remote.Identifier parses the reponame accurately

ashwin-venkatesh:
Signed-off-by: Ashwin Venkatesh <avenkatesh@pivotal.io>
Signed-off-by: Matthew McNew <mmcnew@pivotal.io>

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -163,12 +163,17 @@ func (i *Image) Found() bool {
 }
 
 func (i *Image) Identifier() (imgutil.Identifier, error) {
+	ref, err := name.ParseReference(i.repoName, name.WeakValidation)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


jromero:
What were the cases in which it wasn't parsing it properly? Was it when repo name already had a tag?

ashwin-venkatesh:
> What were the cases in which it wasn't parsing it properly? Was it when repo name already had a tag?

This was an issue when we invoked remote.NewImage() with an image name that had a fully qualified sha256 at the end of it. This would cause the previous Identifier implementation to blow up as it would try to append a SHA at the end of the above SHA ie image/foo@sha256lakdsjfhaks@sha256aalskdfasl

This implementation will trim the sha from the repo name in case it does exist

PR #16, opened 2019-10-18
Use default compression for remote image

### LINKED ISSUE #15: Use default compression for remote image AddLayer
Based on discoveries mentioned here (https://github.com/buildpack/pack/pull/351#discussion_r336084550), we need to use `gzip.DefaultCompression` when [creating a `tarball.Layer`](https://github.com/buildpack/imgutil/blob/cff0c00202d01c688dcecf05bed6bbb5e91249e9/remote/remote.go#L308) to simulate `docker push` compression and layer digest generation.

This could easily be addressed when https://github.com/google/go-containerregistry/pull/574 is merged or if needed we could compress the tar file ourselves and use `tarball.LayerFromOpener`.

jromero:
* Update vendor
* Fix: tagging of initial image with updated docker version

Resolves #15


local/local.go
```diff
@@ -316,8 +316,8 @@ func (i *Image) Save(additionalNames ...string) error {
 	i.inspect = inspect
 
 	var errs []imgutil.SaveDiagnostic
-	for _, n := range additionalNames {
-		if err := i.docker.ImageTag(context.Background(), i.repoName, n); err != nil {
+	for _, n := range append([]string{i.Name()}, additionalNames...) {
```
ameyer-pivotal: This is one of the main changes we had to make when the ggcr update updated docker. For some reason, adding the original image name to `RepoTags` in the manifest no longer results in the initial image (saved during `doSave()`) getting tagged, so we include the original name here when processing additional tags. We can dig in deeper later to understand better what docker requires during save, but this approach works equally as well as before.

PR #20, opened 2020-01-13
Ensure all required configfile fields are set to reasonable values

ekcasey:
* Current defaults won't work for windows but this was already true
* Fixes missing arch in local images
* Fixes empty os/arch in images built from scratch
* reorgs test files to be symmetrical

### DIFF
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -51,40 +51,53 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 	})
 
 	when("#NewImage", func() {
-		when("image is available locally", func() {
-			var repoName = newTestImageName()
-
-			it.After(func() {
-				h.AssertNil(t, h.DockerRmi(dockerClient, repoName))
+		when("no base image is given", func() {
+			it("returns an empty image", func() {
+				_, err := local.NewImage(newTestImageName(), dockerClient)
+				h.AssertNil(t, err)
 			})
 
-			it("returns the local image", func() {
-				labels := make(map[string]string)
-				labels["some.label"] = "some.value"
+			it("sets sensible defaults for all required fields", func() {
+				// os, architecture, and rootfs are required per https://github.com/opencontainers/image-spec/blob/master/config.md
+				img, err := local.NewImage(newTestImageName(), dockerClient)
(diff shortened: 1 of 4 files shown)


local/local_test.go
```diff
@@ -51,40 +51,53 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertEq(t, inspect.Os, "linux")
+				h.AssertEq(t, inspect.Architecture, "amd64")
+				h.AssertEq(t, inspect.RootFS.Type, "layers")
 			})
+		})
 
-			it("returns the local image", func() {
-				labels := make(map[string]string)
-				labels["some.label"] = "some.value"
+		when("#FromBaseImage", func() {
```
ameyer-pivotal: This looks to be accidentally nested under `#NewImage`

local/local_test.go
```diff
@@ -51,40 +51,53 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
+			when("base image exists", func() {
+				var repoName = newTestImageName()
 
-				h.CreateImageOnLocal(t, dockerClient, repoName, fmt.Sprintf(`
+				it.After(func() {
+					h.AssertNil(t, h.DockerRmi(dockerClient, repoName))
```
ameyer-pivotal: It seems strange to have this here without a `Before`, since the image is created in the test. Maybe move it to the end of the test? I suppose it might not get called on a failure though...

I know we discussed the usage of `Before` and `After` previously. In this case, maybe it's best to just put the creation in a `Before` until we have a better pattern for these scenarios.

local/local_test.go
ekcasey: That is because `FromBaseImage` is a option that is provided to `NewImage`

local/local_test.go
ekcasey: In the test I added I actually did the cleanup in a `defer` statement. I just moved this text into the appropriate context so that the organization would mirror that of of `remote_test.go`

ameyer-pivotal:
@ekcasey How important do you think it is that `pack` pick up these changes by bumping imgutil?

PR #24, opened 2020-03-09
Fix local valid name/tag generation in local

micahyoung:
Local `image.Save()` was accepting incomplete image names and relying on the daemon to append the `latest` tag names, while ignoring an error in the response:

```json
{"errorDetail":{"message":"invalid tag \"localhost:32808/pack-image-test-rtgkvgvshf\""},"error":"invalid tag \"localhost:32808/pack-image-test-rtgkvgvshf\""}
```

This PR changes `Save()` to use ggcr `tag.Name()` to explicitly append `latest` tag for incomplete names and additionally adds error checking to catch this type of daemon error from `dockerClient.ImageLoad()`.

Commit details:

- Use ggcr tag.Name() to append 'latest' when given just a name without a tag. Previously, this action was done by undocumented docker.ImageLoad behavior, actually performed on the daemon which would return an error message in the response body that was ignored (not a Golang err).
- Note: the CLI with `docker image load` does not show this undocumented behavior: it fails and does not save when loading an image tarball with a manifest containing a `RepoTags` entry with a name but no explicit tag (e.g. `alpine` instead of `alpine:latest`)
- Add checking for daemon error messages. Previously, when given a name without a tag, the daemon would return an 'invalid tag' error in the response body, yet surprisingly would still write the image, appending 'latest' tag to its saved image name. This func was ignoring the embedded error in the response body and continued as if it were successful.
- Add missing error check for the channel used by the go routine calling `docker.ImageLoad`. Previously, this had been ignored, none of our tests had a scenario where it would have been set. Note: the `ImageLoad` error described above would *not* have returned a Golang error, instead only setting an embedded error in the response.
- Add vendored dependencies from `docker/pkg/jsonmessage`


micahyoung:
Note: this came out of the Windows image/daemon work but I wanted to see what you all think independently, so it didn't get lost.

local/local.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/docker/docker/pkg/jsonmessage"
+
```
ekcasey: can we group this with the other external imports below?

micahyoung: Done!

local/local.go
```diff
@@ -351,11 +354,14 @@ func (i *Image) doSave() (types.ImageInspect, error) {
...
 		}
+
+		if err := checkResponseError(res.Body); err != nil {
+			done <- err // add error but continue, to ensure reader is closed
+		}
+
 		if err := ensureReaderClosed(res.Body); err != nil {
 			done <- errors.Wrap(err, "failed to drain and close response from docker client")
-			return
 		}
```
micahyoung: I rebased against the changes from #28, and I hope this is then intended behavior - adding any `res.Body` errors to the channel but deferring return until `ensureReaderClosed` can flush and close the rest of the response. Look ok?

local/local.go
```diff
@@ -351,11 +354,15 @@ func (i *Image) doSave() (types.ImageInspect, error) {
...
+
+		if err := checkResponseError(res.Body); err != nil {
+			done <- err // add error but continue, to ensure reader is closed
+		}
+
 		if err := ensureReaderClosed(res.Body); err != nil {
 			done <- errors.Wrap(err, "failed to drain and close response from docker client")
 			return
 		}
```
micahyoung: I rebased against the changes from #28, and I hope this is the now- intended behavior - adding any `res.Body` errors to the channel but deferring return until `ensureReaderClosed` can flush and close the rest of the response. Look ok?

local/local_test.go
```diff
@@ -927,131 +927,158 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
+
+				img, err := local.NewImage(repoName, dockerClient)
+				h.AssertNil(t, err)
+
+				err = img.AddLayer(invalidLayerPath)
+				h.AssertNil(t, err)
+
+				err = img.Save(repoName)
+				h.AssertError(t, err, fmt.Sprintf("failed to write image to the following tags: [%s:", repoName))
+				h.AssertError(t, err, "daemon response")
```
micahyoung: Here's my attempt at some test coverage for this. I'm not sure how to repro the unclosed-response-body situation but this at least covers the sad-path of the daemon returning an error message but no golang `error`

local/local.go
ekcasey: Sending on the `done` chan will unblock the next interaction with the docker client (`i.docker.ImageInspectWithRaw` on line 423). We need to flush and close the response before sending another request to the daemon. We could save the error to return after we ensureReaderClose (instead of the close error if there is one).

micahyoung: Oh cool. I'll try that then! Thanks!

local/local.go
```diff
@@ -351,10 +354,17 @@ func (i *Image) doSave() (types.ImageInspect, error) {
...
+		//only return response error after response is drained and closed
+		responseErr := checkResponseError(res.Body)
+		drainCloseErr := ensureReaderClosed(res.Body)
+		if responseErr != nil {
+			done <- responseErr
 			return
 		}
+		if drainCloseErr != nil {
+			done <- drainCloseErr
+		}
```
micahyoung: Ok, here's another attempt. Is it ok that we're explicitly prioritizing response error over the close error? My feeling is that the response errors would be important but I haven't seen many examples of `io.Copy()` or `r.Close()` errors.

ekcasey: I think prioritizing the response error is the right call.

PR #28, opened 2020-03-11
ensure we read complete response from docker

ekcasey:
* so that the docker client can reuse the same connection, we must fully drain every response before making another request

Signed-off-by: Emily Casey <ecasey@pivotal.io>

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -351,8 +351,10 @@ func (i *Image) doSave() (types.ImageInspect, error) {
 			done <- err
 			return
 		}
-		defer res.Body.Close()
-		io.Copy(ioutil.Discard, res.Body)
+		if err := ensureReaderClosed(res.Body); err != nil {
+			done <- errors.Wrap(err, "failed to drain and close response from docker client")
+			return
+		}
 
 		done <- nil
 	}()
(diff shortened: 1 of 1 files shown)


local/local.go
```diff
@@ -683,3 +685,10 @@ func v1Config(inspect types.ImageInspect) (v1.ConfigFile, error) {
 		Config: config,
 	}, nil
 }
+
+func ensureReaderClosed(r io.ReadCloser) error {
+	if _, err := io.Copy(ioutil.Discard, r); err !=nil {
+		return err
+	}
+	return r.Close()
```
zmackie: This is a different behavior in handling the close. If the defer call above errored, we'd just be dropping that on the floor. This seems better, but just wanted to call this change out.

zmackie:
Not that it matters :) , but LGTM.

local/local.go
ekcasey: good catch! fixed

local/local.go
```diff
@@ -351,8 +351,10 @@ func (i *Image) doSave() (types.ImageInspect, error) {
 			done <- err
 			return
 		}
-		defer res.Body.Close()
-		io.Copy(ioutil.Discard, res.Body)
+		if err := ensureReaderClosed(res.Body); err != nil {
```
jromero: This solely seems to be discarding the response. How would that work or should this be changed to account for #24?

ekcasey: We were already discarding the response on master. The important thing here is to finish reading the entire response and closing before we send on the `done` chan and unblock the next interaction with the docker client (if the reader is not closed the client can't reuse the same connection).

This will have to be thoughtfully merged with #24 to avoid reintroducing the bug.

local/local.go
```diff
@@ -683,3 +685,13 @@ func v1Config(inspect types.ImageInspect) (v1.ConfigFile, error) {
 		Config: config,
 	}, nil
 }
+
+// ensureReaderClosed drains and closes and reader, returning the first error
+func ensureReaderClosed(r io.ReadCloser) error {
+	var err error
+	_, err = io.Copy(ioutil.Discard, r)
```
sclevine: `_, err :=`

sclevine: (unless the separated declaration was stylistic and not the result of refactoring)

ekcasey: fixed

PR #33, opened 2020-03-23
Refactor tests not rely on Dockerfile FROM scratch

micahyoung:
- Also Replace some "FROM busybox" examples that didn't need "RUN"
- Purpose: changed to ease porting to Windows since Windows daemons do not support FROM scratch
- Benefit: Tests are less coupled to Dockerfile and run faster
- Potential downsides: tests now depend more on imgutil methods instead of Dockerfiles and docker daemon impl to create test images but it's unlikely Dockerfile and imgutil functionality will be out of sync in a way that shouldn't be fixed as a bug

Note: in upcoming Windows PRs, the remaining `FROM busybox` that have `RUN <command>` will get cleaned up as well, but the cleanup doesn't make much sense in this stand-alone commit.

### DIFF
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -3,8 +3,6 @@ package local_test
 import (
 	"archive/tar"
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -629,19 +627,15 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 			repoName = newTestImageName()
 		)
 		it.Before(func() {
+			var err error
 			h.CreateImageOnLocal(t, dockerClient, repoName, fmt.Sprintf(`
 				FROM busybox
 				LABEL repo_name_for_randomisation=%s
 				RUN echo -n old-layer > old-layer.txt
 			`, repoName), nil)
-			tr, err := h.CreateSingleFileTar("/new-layer.txt", "new-layer")
+
+			tarPath, err = h.CreateSingleFileTar("/new-layer.txt", "new-layer")
 			h.AssertNil(t, err)
-			tarFile, err := ioutil.TempFile("", "add-layer-test")
-			h.AssertNil(t, err)
-			defer tarFile.Close()
-			_, err = io.Copy(tarFile, tr)
-			h.AssertNil(t, err)
-			tarPath = tarFile.Name()
 
 			img, err = local.NewImage(
 				repoName, dockerClient,
@@ -683,22 +677,19 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 			repoName = newTestImageName()
 		)
 		it.Before(func() {
+			var err error
+
 			h.CreateImageOnLocal(t, dockerClient, repoName, fmt.Sprintf(`
 				FROM busybox
 				LABEL repo_name_for_randomisation=%s
 				RUN echo -n old-layer > old-layer.txt
 			`, repoName), nil)
-			tr, err := h.CreateSingleFileTar("/new-layer.txt", "new-layer")
+
+			tarPath, err = h.CreateSingleFileTar("/new-layer.txt", "new-layer")
 			h.AssertNil(t, err)
-			tarFile, err := ioutil.TempFile("", "add-layer-test")
+
+			diffID, err = h.FileDiffID(tarPath)
 			h.AssertNil(t, err)
-			defer tarFile.Close()
-			hasher := sha256.New()
-			mw := io.MultiWriter(tarFile, hasher)
-			_, err = io.Copy(mw, tr)
-			h.AssertNil(t, err)
-			tarPath = tarFile.Name()
-			diffID = "sha256:" + hex.EncodeToString(hasher.Sum(make([]byte, 0, hasher.Size())))
 
 			img, err = local.NewImage(
 				repoName, dockerClient,
@@ -706,6 +697,7 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 				local.WithPreviousImage(repoName),
 			)
 			h.AssertNil(t, err)
+
 			origID = h.ImageID(t, repoName)
 		})
 
@@ -904,21 +896,12 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 				img, err = local.NewImage(repoName, dockerClient)
 				h.AssertNil(t, err)
 
-				tr, err := h.CreateSingleFileTar("/new-layer.txt", "new-layer")
+				tarPath, err = h.CreateSingleFileTar("/new-layer.txt", "new-layer")
 				h.AssertNil(t, err)
-
-				tarFile, err := ioutil.TempFile("", "add-layer-test")
-				h.AssertNil(t, err)
-				defer tarFile.Close()
-
-				_, err = io.Copy(tarFile, tr)
-				h.AssertNil(t, err)
-
-				tarPath = tarFile.Name()
 			})
 
 			it.After(func() {
-				os.Remove(tarPath)
+				h.AssertNil(t, os.Remove(tarPath))
 				h.AssertNil(t, h.DockerRmi(dockerClient, repoName, origID))
 			})
 
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -75,14 +75,15 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 
 		when("#FromBaseImage", func() {
 			when("base image exists", func() {
+				var baseImageName = newTestImageName()
 				var repoName = newTestImageName()
 
 				it.After(func() {
-					h.AssertNil(t, h.DockerRmi(dockerClient, repoName))
+					h.AssertNil(t, h.DockerRmi(dockerClient, baseImageName))
 				})
 
 				it("returns the local image", func() {
-					baseImage, err := local.NewImage(repoName, dockerClient)
+					baseImage, err := local.NewImage(baseImageName, dockerClient)
 					h.AssertNil(t, err)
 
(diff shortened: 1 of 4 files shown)


local/local_test.go
```diff
@@ -133,11 +132,13 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
 			it.Before(func() {
-				h.CreateImageOnLocal(t, dockerClient, repoName, fmt.Sprintf(`
-					FROM scratch
-					LABEL repo_name_for_randomisation=%s
-					LABEL mykey=myvalue other=data
-				`, repoName), nil)
+				baseImage, err := local.NewImage(repoName, dockerClient)
+				h.AssertNil(t, err)
+
+				h.AssertNil(t, baseImage.SetLabel("repo_name_for_randomisation", repoName))
```
micahyoung: Is this label only needed for cache-busting on the daemon? Do we know if these are still needed?

remote/remote_test.go
```diff
@@ -537,10 +540,11 @@ func testRemoteImage(t *testing.T, when spec.G, it spec.S) {
 		)
 		it.Before(func() {
 			h.CreateImageOnRemote(t, dockerClient, repoName, fmt.Sprintf(`
```
ekcasey: same comment as before, can we move away from Dockerfiles entirely?

remote/remote_test.go
```diff
@@ -703,11 +708,16 @@ func testRemoteImage(t *testing.T, when spec.G, it spec.S) {
 			var tarPath string
 
 			it.Before(func() {
-				h.CreateImageOnRemote(t, dockerClient, repoName, fmt.Sprintf(`
-					FROM busybox
-					LABEL repo_name_for_randomisation=%s
-					LABEL mykey=oldValue
-				`, repoName), nil)
+				baseImage, err := remote.NewImage(
```
ekcasey: unused `baseImage`?

local/local_test.go
ekcasey: If we make this change we should be careful that we aren't relying on Force to paper over incomplete cleanup, to avoid leaking images.

local/local_test.go
```diff
@@ -436,10 +443,12 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
-					FROM scratch
-					LABEL repo_name_for_randomisation=%s
-				`, repoName), nil)
+			baseImage, err := local.NewImage(repoName, dockerClient)
+			h.AssertNil(t, err)
+
+			h.AssertNil(t, baseImage.SetLabel("repo_name_for_randomisation", repoName))
+			h.AssertNil(t, baseImage.Save())
+
 			img, err = local.NewImage(repoName, dockerClient)
```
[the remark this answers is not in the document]
micahyoung: The old behavior seemed to be that the Dockerfile was used to created a daemon image with the name `repoName`, then `local.NewImage` makes an image instance at the same name and eventually clobbers the old one. I'm not sure why this was there (maybe to make cleanup easier) but I kept the behavior for consistency and not wanting to lose test coverage. If we're ok with going away from that, I'll do another pass to remove the name-clobbering test pattern. I'll make sure there's at least one test (and I think there's already one) that ensures this clobbering behavior continues to work.

local/local_test.go
```diff
@@ -630,10 +643,10 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 			)
 			it.Before(func() {
 				h.CreateImageOnLocal(t, dockerClient, repoName, fmt.Sprintf(`
```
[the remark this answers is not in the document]
micahyoung: Yeah, I was hoping we could do that! Should be a quick change and will simplify some upcoming Windows tests. Do we feel there's any concerns having less coverage for `Dockerfile RUN` generated layers?

local/local_test.go
micahyoung: I did another pass but kept the `repo_name_for_randomisation` labels for now, as did I keep `busybox` as a base image since I don't quite understand the tests well enough to be sure what the requirements are for both. I might need a pair to get all the way there.

local/local_test.go
```diff
@@ -401,11 +406,13 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
-			h.CreateImageOnLocal(t, dockerClient, repoName, fmt.Sprintf(`
-					FROM scratch
-					LABEL repo_name_for_randomisation=%s
-					LABEL some-key=some-value
-				`, repoName), nil)
+			baseImage, err := local.NewImage(repoName, dockerClient)
+			h.AssertNil(t, err)
+
+			h.AssertNil(t, baseImage.SetLabel("repo_name_for_randomisation", repoName))
+			h.AssertNil(t, baseImage.SetLabel("some-key", "some-value"))
```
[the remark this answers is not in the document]
micahyoung: Fixed

local/local_test.go
```diff
@@ -401,11 +406,13 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
+			h.AssertNil(t, baseImage.Save())
+
 			img, err = local.NewImage(repoName, dockerClient)
```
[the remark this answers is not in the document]
micahyoung: Fixed by using different images names instead of relying on the same one

local/local_test.go
```diff
@@ -75,23 +72,23 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
-							LABEL repo_name_for_randomisation=%s
-							ENV MY_VAR=my_val
-							`, repoName), labels)
+					h.AssertNil(t, baseImage.SetLabel("repo_name_for_randomisation", repoName))
+					h.AssertNil(t, baseImage.SetEnv("MY_VAR", "my_val"))
+					h.AssertNil(t, baseImage.SetLabel("some.label", "some.value"))
+					h.AssertNil(t, baseImage.Save())
 
-					localImage, err := local.NewImage(repoName, dockerClient, local.FromBaseImage(repoName))
+					localImage, err := local.NewImage(repoName, dockerClient, local.FromBaseImage(baseImageName))
```
micahyoung: Use separate `baseImageName` and `repoName` instead clobbering `repoName`

local/local_test.go
```diff
@@ -133,11 +130,13 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 			var repoName = newTestImageName()
 
 			it.Before(func() {
-				h.CreateImageOnLocal(t, dockerClient, repoName, fmt.Sprintf(`
-					FROM scratch
-					LABEL repo_name_for_randomisation=%s
-					LABEL mykey=myvalue other=data
-				`, repoName), nil)
+				existingImage, err := local.NewImage(repoName, dockerClient)
```
micahyoung: Using `existingImage` instead of `baseImage` (in earlier commit) to match the context and clarify that this is about images that already exist

local/local_test.go
```diff
@@ -394,29 +383,19 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
-			var err error
-			h.CreateImageOnLocal(t, dockerClient, repoName, fmt.Sprintf(`
-					FROM scratch
-					LABEL repo_name_for_randomisation=%s
-					LABEL some-key=some-value
-				`, repoName), nil)
-			img, err = local.NewImage(repoName, dockerClient)
-			h.AssertNil(t, err)
-			origID = h.ImageID(t, repoName)
-		})
```
micahyoung: Remove unneeded setup

local/local_test.go
```diff
@@ -540,42 +491,71 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
-				go func() {
-					defer wg.Done()
-					newBase = "pack-newbase-test-" + h.RandString(10)
-					h.CreateImageOnLocal(t, dockerClient, newBase, fmt.Sprintf(`
-						FROM busybox
-						LABEL repo_name_for_randomisation=%s
-						RUN echo new-base > base.txt
-						RUN echo text-new-base > otherfile.txt
-					`, newBase), nil)
-				}()
```
micahyoung: WaitGroup was no longer needed since creation trivially fast

local/local_test.go
```diff
@@ -540,42 +491,71 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
...
+
+				imgLayer2Path, err := h.CreateSingleFileTar("/myimage2.txt", "text-from-image")
+				h.AssertNil(t, err)
+				defer os.Remove(imgLayer2Path)
+
+				h.AssertNil(t, origImage.AddLayer(imgLayer1Path))
+				h.AssertNil(t, origImage.AddLayer(imgLayer2Path))
+
+				h.AssertNil(t, origImage.Save())
+
```
micahyoung: There are more lines here. Should we reduce this test case to 1 layer per image?

remote/remote_test.go
```diff
@@ -285,6 +282,15 @@ func testRemoteImage(t *testing.T, when spec.G, it spec.S) {
...
 			it("returns the new digest reference", func() {
+				var err error
+
+				img, err := remote.NewImage(
+					repoName+":some-tag",
+					authn.DefaultKeychain,
+					remote.FromBaseImage("busybox@sha256:915f390a8912e16d4beb8689720a17348f3f6d1a7b659697df850ab625ea29d5"),
+				)
+				h.AssertNil(t, err)
+
```
micahyoung: Duplicate image instantiation to avoid existing complex (and incorrect) nesting.

local/local_test.go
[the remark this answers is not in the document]
micahyoung: Now disambiguated in subsequent commit.

local/local_test.go
```diff
@@ -924,7 +925,15 @@ func testLocalImage(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		it("does not download the old image if layers are directly above (performance)", func() {
```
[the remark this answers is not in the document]
micahyoung: Removed the test

PR #42, opened 2020-04-08
Add support for Windows images

micahyoung:
The intent is to enable imgutil consumers to create Windows and Linux OCI images. 

For local, a Windows image is automatically created if the daemon is Windows, and the Architecture and OS Version will also match. If it's a Linux daemon, the empty image will be Linux. This can be overridden by a base image if desired but will fail to save if the base image OS, OS Version, or Architecture is incompatible with the daemon.

For remote, new images will still always default to `linux/amd64`. Any compliant registry should be able to save any type of valid image, Windows or Linux. 

Change summary:
- remote: NewImage: Use ggcr's default platform preferences (`linux/amd64`) but this is overridden by any base images from `remote.FromBaseImage` option.
- local: explicitly set image's default platform to match the daemon OS/OSVersion/Arch for empty images since only matching images can be stored on the daemon. Note: non-matching architectures (e.g `arm64`) are potentially allowed on a daemon (e.g. `amd64`).
- tests: add coverage for new behavior scenarios
- tests: all tests should be runnable on Linux or Mac GOOS + Docker with Linux containers OR Windows GOOS (1809 or greater) + Docker with Windows containers
- tests: replace "busybox" with specific image manifest references for either Linux or Windows.
- tests: replace "registry:2" with compatible registry container images for Linux or Windows
- tests: change helpers to generate Windows specific layer requirements: base images need specific proprietary files, upper layers need specific parent directories
- tests: use non-default container process isolation to speed up Windows tests
- github actions: add identical jobs for windows

Signed-off-by: Andrew Meyer <meyeran@vmware.com>

### DIFF
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -61,6 +61,14 @@ func (i *Image) Label(key string) (string, error) {
 	return i.labels[key], nil
 }
 
+func (i *Image) OS() (string, error) {
+	return "linux", nil
+}
+
+func (i *Image) Architecture() (string, error) {
+	return "amd64", nil
+}
+
 func (i *Image) Rename(name string) {
 	i.name = name
 }
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -31,6 +31,7 @@ func NewImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
 		createdAt:     time.Now(),
 		savedNames:    map[string]bool{},
 		os:            "linux",
+		osVersion:     "",
 		architecture:  "amd64",
 	}
 }
@@ -67,11 +68,15 @@ func (i *Image) Label(key string) (string, error) {
 }
 
 func (i *Image) OS() (string, error) {
-	return "linux", nil
+	return i.os, nil
+}
+
+func (i *Image) OSVersion() (string, error) {
+	return i.os, nil
 }
 
 func (i *Image) Architecture() (string, error) {
-	return "amd64", nil
+	return i.architecture, nil
 }
 
 func (i *Image) Rename(name string) {
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -30,6 +30,8 @@ func NewImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
 		prevLayersMap: map[string]string{},
 		createdAt:     time.Now(),
 		savedNames:    map[string]bool{},
+		os:            "linux",
+		architecture:  "amd64",
 	}
 }
 
@@ -42,6 +44,9 @@ type Image struct {
 	labels        map[string]string
 	env           map[string]string
 	topLayerSha   string
+	os            string
+	osVersion     string
+	architecture  string
 	identifier    imgutil.Identifier
 	name          string
 	entryPoint    []string
@@ -239,6 +244,12 @@ func (i *Image) Found() bool {
 
 func (i *Image) SetIdentifier(identifier imgutil.Identifier) {
 	i.identifier = identifier
+}
+
+func (i *Image) SetPlatform(os, osVersion, architecture string) {
+	i.os = os
+	i.osVersion = osVersion
+	i.architecture = architecture
 }
 
 func (i *Image) Cleanup() error {
--- a/local/local.go
+++ b/local/local.go
@@ -75,7 +75,12 @@ func FromBaseImage(imageName string) ImageOption {
 }
 
 func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...ImageOption) (imgutil.Image, error) {
-	inspect := defaultInspect()
+	var err error
+
+	inspect, err := defaultInspect(dockerClient)
+	if err != nil {
+		return nil, err
+	}
 
 	image := &Image{
 		docker:       dockerClient,
@@ -85,7 +90,6 @@ func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...Image
 		downloadOnce: &sync.Once{},
 	}
 
-	var err error
 	for _, v := range ops {
 		image, err = v(image)
 		if err != nil {
@@ -109,6 +113,14 @@ func (i *Image) Env(key string) (string, error) {
 		}
 	}
 	return "", nil
+}
+
+func (i *Image) OS() (string, error) {
+	return i.inspect.Os, nil
+}
+
+func (i *Image) Architecture() (string, error) {
+	return i.inspect.Architecture, nil
 }
 
 func (i *Image) Rename(name string) {
@@ -608,7 +620,7 @@ func inspectOptionalImage(docker client.CommonAPIClient, imageName string) (type
 
 	if inspect, _, err = docker.ImageInspectWithRaw(context.Background(), imageName); err != nil {
 		if client.IsErrNotFound(err) {
-			return defaultInspect(), nil
+			return types.ImageInspect{}, nil
 		}
 
 		return types.ImageInspect{}, errors.Wrapf(err, "verifying image '%s'", imageName)
@@ -617,12 +629,18 @@ func inspectOptionalImage(docker client.CommonAPIClient, imageName string) (type
 	return inspect, nil
 }
 
-func defaultInspect() types.ImageInspect {
+func defaultInspect(docker client.CommonAPIClient) (types.ImageInspect, error) {
+	daemonInfo, err := docker.Info(context.Background())
+	if err != nil {
+		return types.ImageInspect{}, err
+	}
+
 	return types.ImageInspect{
-		Os:           "linux",
+		Os:           daemonInfo.OSType,
+		OsVersion:    daemonInfo.OSVersion,
 		Architecture: "amd64",
 		Config:       &container.Config{},
-	}
+	}, nil
 }
 
 func v1Config(inspect types.ImageInspect) (v1.ConfigFile, error) {
@@ -689,6 +707,7 @@ func v1Config(inspect types.ImageInspect) (v1.ConfigFile, error) {
 		Created:      v1.Time{Time: imgutil.NormalizedDateTime},
 		History:      history,
 		OS:           inspect.Os,
+		OSVersion:    inspect.OsVersion,
 		RootFS: v1.RootFS{
 			Type:    "layers",
 			DiffIDs: diffIDs,
--- a/local/local.go
+++ b/local/local.go
@@ -119,6 +119,10 @@ func (i *Image) OS() (string, error) {
 	return i.inspect.Os, nil
 }
 
+func (i *Image) OSVersion() (string, error) {
+	return i.inspect.OsVersion, nil
+}
+
 func (i *Image) Architecture() (string, error) {
 	return i.inspect.Architecture, nil
 }
--- a/local/local.go
+++ b/local/local.go
@@ -620,7 +620,7 @@ func inspectOptionalImage(docker client.CommonAPIClient, imageName string) (type
 
 	if inspect, _, err = docker.ImageInspectWithRaw(context.Background(), imageName); err != nil {
 		if client.IsErrNotFound(err) {
-			return types.ImageInspect{}, nil
+			return defaultInspect(docker)
 		}
 
 		return types.ImageInspect{}, errors.Wrapf(err, "verifying image '%s'", imageName)
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -607,7 +607,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			it("switches the base", func() {
 				// Before
-				txt, err := h.CopySingleFileFromImage(dockerClient, repoName, "/base.txt")
+				txt, err := h.CopySingleFileFromLocalImage(dockerClient, repoName, "/base.txt")
 				h.AssertNil(t, err)
 				h.AssertEq(t, txt, "old-base")
 
@@ -717,7 +717,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				h.AssertNil(t, img.Save())
 				defer h.DockerRmi(dockerClient, repoName)
 
-				output, err := h.CopySingleFileFromImage(dockerClient, repoName, "new-layer.txt")
+				output, err := h.CopySingleFileFromLocalImage(dockerClient, repoName, "new-layer.txt")
 				h.AssertNil(t, err)
 				h.AssertEq(t, output, "new-layer")
 			})
@@ -758,11 +758,11 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				h.AssertNil(t, img.Save())
 				defer h.DockerRmi(dockerClient, repoName)
 
-				output, err := h.CopySingleFileFromImage(dockerClient, repoName, "old-layer.txt")
+				output, err := h.CopySingleFileFromLocalImage(dockerClient, repoName, "old-layer.txt")
 				h.AssertNil(t, err)
 				h.AssertEq(t, output, "old-layer")
 
-				output, err = h.CopySingleFileFromImage(dockerClient, repoName, "new-layer.txt")
+				output, err = h.CopySingleFileFromLocalImage(dockerClient, repoName, "new-layer.txt")
 				h.AssertNil(t, err)
 				h.AssertEq(t, output, "new-layer")
 			})
@@ -817,11 +817,11 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 			h.AssertNil(t, img.AddLayerWithDiffID(newLayerPath, diffID))
 			h.AssertNil(t, img.Save())
 
-			output, err := h.CopySingleFileFromImage(dockerClient, repoName, "old-layer.txt")
+			output, err := h.CopySingleFileFromLocalImage(dockerClient, repoName, "old-layer.txt")
 			h.AssertNil(t, err)
 			h.AssertEq(t, output, "old-layer")
 
-			output, err = h.CopySingleFileFromImage(dockerClient, repoName, "new-layer.txt")
+			output, err = h.CopySingleFileFromLocalImage(dockerClient, repoName, "new-layer.txt")
 			h.AssertNil(t, err)
 			h.AssertEq(t, output, "new-layer")
 		})
@@ -965,12 +965,12 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			h.AssertNil(t, img.Save())
 
-			output, err := h.CopySingleFileFromImage(dockerClient, repoName, "layer-2.txt")
+			output, err := h.CopySingleFileFromLocalImage(dockerClient, repoName, "layer-2.txt")
 			h.AssertNil(t, err)
 			h.AssertEq(t, output, "old-layer-2")
 
 			// Confirm layer-1.txt does not exist
-			_, err = h.CopySingleFileFromImage(dockerClient, repoName, "layer-1.txt")
+			_, err = h.CopySingleFileFromLocalImage(dockerClient, repoName, "layer-1.txt")
 			h.AssertMatch(t, err.Error(), regexp.MustCompile(`Error: No such container:path: .*:layer-1.txt`))
 		})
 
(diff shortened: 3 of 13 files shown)


ameyer-pivotal:
Noticed some breakages in downstream repos like `pack` due to the fake image not implementing `imgutil.Image` anymore. We should fix this here.

remote/remote.go
```diff
@@ -53,7 +58,34 @@ func WithPreviousImage(imageName string) ImageOption {
 func FromBaseImage(imageName string) ImageOption {
 	return func(r *Image) (*Image, error) {
 		var err error
-		r.image, err = newV1Image(r.keychain, imageName)
+
+		currentImagePlatform, err := imagePlatform(r.image)
```
ekcasey, line 62: It looks like `WithDefaultPlatform` option must be passed before the `FromBaseImage` option to get the correct value here. That might not be obvious to library consumers. I want to change `FromBaseImage` to take an image rather than a string which would make this case easier, but in the meantime we probably need to solve the ordering problem.

We could potentially change the options to produce a config that is then applied or instead of mutating the image direction in option func.

remote/remote.go
```diff
@@ -53,7 +58,34 @@ func WithPreviousImage(imageName string) ImageOption {
...
+
+		r.image, err = newV1Image(r.keychain, imageName, currentImagePlatform)
+		if err != nil {
+			return nil, err
+		}
+		return r, nil
+	}
+}
+
+func WithDefaultPlatform(p v1.Platform) ImageOption {
```
ekcasey, line 75: Right now we try not expose ggcr types directly in the public interface of imgutil (`authn.Keychain` is the convenient exception). I think we should add our own Platform type.

remote/remote.go
```diff
@@ -35,7 +35,12 @@ func WithPreviousImage(imageName string) ImageOption {
 	return func(r *Image) (*Image, error) {
 		var err error
 
-		prevImage, err := newV1Image(r.keychain, imageName)
+		currentImagePlatform, err := imagePlatform(r.image)
```
ekcasey, line 38: same ordering constraints introduced here

testhelpers/testhelpers.go
```diff
@@ -136,11 +144,16 @@ func Eventually(t *testing.T, test func() bool, every time.Duration, timeout tim
 	}
 }
 
-func PullImage(dockerCli dockercli.CommonAPIClient, ref string) error {
-	rc, err := dockerCli.ImagePull(context.Background(), ref, dockertypes.ImagePullOptions{})
+func PullImage(dockerCli dockercli.CommonAPIClient, ref string, registryAuth ...string) error {
```
ekcasey, line 147: It doesn't look like any of the tests are actually passing a registryAuth

testhelpers/testhelpers.go
```diff
@@ -209,8 +235,13 @@ func CopySingleFileFromImage(dockerCli dockercli.CommonAPIClient, repoName, path
 	return CopySingleFileFromContainer(dockerCli, ctrID, path)
 }
 
-func PushImage(dockerCli dockercli.CommonAPIClient, ref string) error {
-	rc, err := dockerCli.ImagePush(context.Background(), ref, dockertypes.ImagePushOptions{RegistryAuth: "{}"})
+func PushImage(dockerCli dockercli.CommonAPIClient, ref string, registryAuth ...string) error {
```
ekcasey, line 238: It doesn't look like any of the tests are actually passing a registryAuth

testhelpers/testhelpers.go
```diff
@@ -243,8 +274,24 @@ func ImageID(t *testing.T, repoName string) string {
...
 }
 
-func CreateSingleFileTar(path, txt string) (string, error) {
-	tarFile, err := ioutil.TempFile("", "create-single-file-tar-path")
+type LayerOption string
+
+const BaseLayerOption LayerOption = "baselayer"
+const WindowsOSOption LayerOption = "windows"
+
+func CreateSingleFileLayerTar(path, txt string, opts ...LayerOption) (string, error) {
```
ekcasey, line 282: The name `CreateSingleFileLayerTar` is a bit misleading when `BaseLayerOption` is passed. Instead of reusing this layer generating helper, maybe we could create a helper for generating base images or something like that.

ekcasey: Alternatively if we can create image (but not a container) w/o a base layer, we could change our tests to stop creating containers just to copy files. It would probably be better to assert on the `inspect.RootFs` to make sure a layer is added, instead of copying a file from the layer out of a container.

remote/remote_test.go
```diff
@@ -720,17 +842,17 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertNil(t, existingImage.SetLabel("mykey", "oldValue"))
+				h.AssertNil(t, existingImage.Save())
 
 				tarFile, err := ioutil.TempFile("", "add-layer-test")
 				h.AssertNil(t, err)
 				defer tarFile.Close()
 
-				tarPath, err = h.CreateSingleFileTar("/new-layer.txt", "new-layer")
+				tarPath, err = h.CreateSingleFileLayerTar("/new-layer.txt", "new-layer")
 				h.AssertNil(t, err)
```
ekcasey, line 856: We don't appear to be using any of this setup logic in the tests.

remote/remote_test.go
```diff
@@ -2,19 +2,17 @@ package remote_test
...
 	"os"
 	"regexp"
-	"strings"
 	"testing"
 	"time"
 
 	"github.com/docker/docker/client"
-	"github.com/google/go-containerregistry/pkg/authn"
+	ggcrauthn "github.com/google/go-containerregistry/pkg/authn"
+	ggcrv1 "github.com/google/go-containerregistry/pkg/v1"
```
ekcasey, line 15: why the aliases?

fakes/image.go
```diff
@@ -61,6 +61,14 @@ func (i *Image) Label(key string) (string, error) {
 	return i.labels[key], nil
 }
 
+func (i *Image) OS() (string, error) {
+	return "linux", nil
+}
+
+func (i *Image) Architecture() (string, error) {
+	return "amd64", nil
+}
```
ekcasey, line 70: We should let users of the fake set these if they wish, instead of hardcoding.

remote/remote.go
```diff
@@ -98,13 +130,43 @@ func newV1Image(keychain authn.Keychain, repoName string) (v1.Image, error) {
...
+			return nil, err
+		}
+	case types.OCIImageIndex, types.DockerManifestList:
+		image, err = remote.Image(ref, remote.WithPlatform(fallbackPlatform), remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport))
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("unknown descriptor response type for ref: %s => %s", repoName, descriptor.MediaType)
+	}
```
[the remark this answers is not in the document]
micahyoung: You're right! I totally missed that (I think I had a red-herring test failure early on)! I'll revert back and just add the platform `remote.WithPlatform(platform)`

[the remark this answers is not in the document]
micahyoung: But all tests do pass with just `remote.Image()`

local/local.go
```diff
@@ -608,7 +620,7 @@ func inspectOptionalImage(docker client.CommonAPIClient, imageName string) (type
 
 	if inspect, _, err = docker.ImageInspectWithRaw(context.Background(), imageName); err != nil {
 		if client.IsErrNotFound(err) {
-			return defaultInspect(), nil
+			return types.ImageInspect{}, nil
```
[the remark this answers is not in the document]
micahyoung: Since the signature for `defaultInspect` changed from `defaultInspect() ImageInspect` to `defaultInspect(dockerClient) (ImageInspect, error)` (with an extra query), I wasn't sure it was worth the extra code for what I (probably naively) assumed was empty result. Happy to add it back though.

local/local_test.go
```diff
@@ -55,15 +68,18 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
 				h.AssertNil(t, err)
 				h.AssertNil(t, img.Save())
+
 				defer h.DockerRmi(dockerClient, img.Name())
 				inspect, _, err := dockerClient.ImageInspectWithRaw(context.TODO(), img.Name())
 				h.AssertNil(t, err)
-				h.AssertEq(t, inspect.Os, "linux")
+
+				// os, architecture come from daemon
+				h.AssertEq(t, inspect.Os, daemonOS)
```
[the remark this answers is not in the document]
micahyoung: Done!

testhelpers/testhelpers.go
micahyoung: Fixed!

remote/remote_test.go
micahyoung: Good catch! I moved the only bit that was used the test that used it.

testhelpers/testhelpers.go
micahyoung: Fixed!

remote/remote.go
micahyoung: Thanks for the feedback. I added `WithDefaultPlatform` without a clear use-case yet, except to exercise the some of the `newV1Image` fallback logic that is now removed. Would it be ok to remove `WithDefaultPlatform` for now, the we can add it back later if there's a need that 'FromBaseImage` doesn't provide?

remote/remote.go
micahyoung: If we remove `WithDefaultPlatform` for now, as mentioned above, then we can punt on this issue for the time being.

remote/remote.go
micahyoung: If we remove `WithDefaultPlatform` for now, then we can have a simpler `fallbackPlatform` and won't need to keep track of the `currentPlatform`

remote/remote_test.go
micahyoung: I'll revert back. The intent was to obviate the leaky `v1.Platform` uses which will now go away if we remove `WithDefaultPlatform`

fakes/image.go
micahyoung: Fixed!

remote/remote.go
ekcasey: I think that makes sense. Generating a valid empty windows image isn't really possible anyways it seems like (you need a base layer with a lot of configuration).

remote/remote_test.go
```diff
@@ -303,97 +419,85 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
 			h.AssertNil(t, img.Save())
 
-			h.AssertNil(t, h.PullImage(dockerClient, repoName))
-			defer h.DockerRmi(dockerClient, repoName)
-
-			inspect, _, err := dockerClient.ImageInspectWithRaw(context.TODO(), repoName)
-			h.AssertNil(t, err)
-
-			h.AssertEq(t, []string(inspect.Config.Cmd), []string{"some", "cmd"})
+			configFile := h.FetchManifestImageConfigFile(t, repoName)
```
natalieparellano, line 499: Just curious, what is the reason for doing away with `dockerClient.ImageInspectWithRaw`?

testhelpers/testhelpers.go
micahyoung: I like the idea of a helper to generate minimal `FROM scratch`-type base images to satisfy whatever daemon. Then we still keep the option of pull/run/push using the `docker.Client`. We had a few other things sprinkled around that could be dried up by that (a default `CMD` comes to mind). I'll do a pass at that here in a new commit.

testhelpers/testhelpers.go
```diff
@@ -243,8 +274,24 @@ func ImageID(t *testing.T, repoName string) string {
 	return inspect.ID
 }
 
-func CreateSingleFileTar(path, txt string) (string, error) {
-	tarFile, err := ioutil.TempFile("", "create-single-file-tar-path")
+type LayerOption string
```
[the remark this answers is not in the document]
micahyoung: The thought was that we'd have a single helper function for creating layers, but you could pass different options to it to get one of three different flavors: regular linux layer, regular windows layer, or windows base layer. After discussion in other comments, it likely makes more sense to combine move the whole base-layer need into an empty-but-valid image helper, which I'm going to make an attempt at in another commit here.

remote/remote_test.go
```diff
@@ -42,75 +40,184 @@ func TestRemote(t *testing.T) {
...
+						img, err := remote.NewImage(
+							repoName,
+							ggcrauthn.DefaultKeychain,
+							remote.FromBaseImage(manifestListName),
+						)
+						h.AssertNil(t, err)
+
+						os, err := img.OS()
+						h.AssertNil(t, err)
+						h.AssertEq(t, os, "linux")
```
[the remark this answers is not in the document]
micahyoung: Yeah, this is where it gets a little wacky. For a minimal constructor call: (say `local.NewImage(`/remote.NewImage(repoName, ggcrauthn.DefaultKeychain`)The `GOOS` would never used, and `remote` *always* defaults to Linux whereas `local` defaults to the Daemon OS.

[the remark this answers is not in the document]
micahyoung: Yeah, this is where it get more subtle than expected. We discovered that `pack` almost always calls `imgutil` with a base image that already has the desired OS/Arch, so all we really need is some valid defaults that will be overridden in most cases. So we can get by without using `GOOS` at all, and have `remote` *always* default to Linux for empty images and have `local` default to the Daemon OS: 

```golang
remote.NewImage("test:tag", authn.DefaultKeychain) // always "linux/amd64"
local.NewImage("test:tag", dockerClient)   // image os/arch == docker os/arch
```

For a `imgutil` consumer (`pack`) to get a `windows` remote image, it would use a base image option when initializing:

```golang
remote.NewImage("test:tag", authn.DefaultKeychain, remote.FromBaseImage("nanoserver:1809"))
local.NewImage("test:tag", dockerClient, local.FromBaseImage("nanoserver:1809"))
```

I was going to add a new option to minimal new remote images (like:  `remote.WithDefaultPlatform("windows", "amd64")`) to get a `windows` image without using a `FromBaseImage` image, but it turns out `pack` wouldn't use it and it just added more complexity around supporting multiple options.

So the only reason either `remote_test.go` needs to know the daemon OS is for a couple tests that create containers for test assertions. If we refactored those tests, `remote_test.go` would be identical for any OS and we wouldn't event need a Daemon to run them (only a registry).

remote/remote_test.go
micahyoung: If we can make all `remote` tests just use direct registry queries (via ggcr) instead of indirectly through `dockerClient.Pull` + `dockerClient.Inspect`, the tests end up being identical for any image OS, run faster, less code and eventually no Daemon dependency.

remote/remote_test.go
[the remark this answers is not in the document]
natalieparellano: Ohhh I see - thank you for the explanation! It might be worth updating the test description [here](https://github.com/buildpacks/imgutil/pull/42/files/d6c8f1ab604eb0b1e8a3075b6c6dcb88f6ed6cbd#diff-f1669c622581f9fccfe59ebda3685db3R162) to capture a bit of the nuance (assuming it doesn't get clobbered in the refactor).

I noticed you said "almost always" about `pack` calling `imgutil` with a base image - are there any edge cases that we need to be worried about?

remote/remote_test.go
natalieparellano: Ah, thank you! That is great!

testhelpers/testhelpers.go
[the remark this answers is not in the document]
natalieparellano: I think this is just me learning how to write effective Go :) I was curious why we needed a new type (instead of just `string`) but now I see that it makes the function definitions much clearer and allows only the defined constants to be passed.

remote/remote_test.go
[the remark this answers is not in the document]
micahyoung: Yeah, the "almost always" for `pack` was for buildpackages which we haven't dug into yet. Presumably those would be "base image-less" images but we're not certain what exactly their needs are yet. There are potentially few different workarounds we could think of that could be performed outside of `imgutil` (by adding a "shim" base image or potentially always storing them as `linux` for remote) but we won't know until we dig in. It still may necessitate an `imgutil` change if those fall short.

testhelpers/testhelpers.go
```diff
@@ -253,17 +300,138 @@ func CreateSingleFileTar(path, txt string) (string, error) {
...
+	}
+
+	if _, err := tw.Write([]byte(txt)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func writeTarSingleFileWindowsBaseLayer(tw *tar.Writer, containerPath, txt string) error {
```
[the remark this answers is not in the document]
micahyoung: Good call. I added a more thorough comment in the latest commit

testhelpers/testhelpers.go
micahyoung: I ended up struggling a little bit with a good helper as I really liked the comprehensibility of the basic `.NewImage(<basic deps>)` calls in the tests and any helper that spits out valid base images would mean we'd have to change them to `.NewImage(<basic deps>, .FromBaseImage(validBase))`. I also tried a helper that takes a new-potentially-invalid image and makes it valid by adding base layers but those felt too magic too. I'll tag the more straightforward, yet more verbose approach I landed on.

remote/remote_test.go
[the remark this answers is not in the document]
natalieparellano: This makes sense, thank you for clarifying!

testhelpers/testhelpers.go
```diff
@@ -253,19 +339,183 @@ func CreateSingleFileTar(path, txt string) (string, error) {
...
+		writeFunc = writeTarSingleFileWindows
+	}
+
+	err = writeFunc(tw, layerPath, txt)
+	if err != nil {
 		return "", err
 	}
 
 	return tarFile.Name(), nil
 }
```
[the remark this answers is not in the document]
ekcasey: The name `CreateSingleFileBaseLayerTar` still throws me, b/c in the windows case there is way more than one file in that layer. Could we just inspect and assert on the contents of `RootFS` or build on top of the `runnableBaseImage` in any test cases where it truly must be runnable (if there are any), and avoid this problem all together?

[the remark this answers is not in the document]
ekcasey: For historical reasons I think these tests are jumping through hoops to copy a file out of layer (in both the daemon and registry) cases when that is not necessary. I think we can trust that if a layer digest is the same, the contents are the same. And in some ways asserting on the contents of a specific file is a less strong assertion.

remote/remote_test.go
```diff
@@ -64,46 +53,108 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+					h.AssertNil(t, err)
+					h.AssertEq(t, arch, "amd64")
+
 					readCloser, err := img.GetLayer(existingLayerSha)
 					h.AssertNil(t, err)
 					defer readCloser.Close()
 				})
+
+				when("base image is a multi-OS/Arch manifest list", func() {
+					it("returns a base image matching the runtime GOOS/GOARCH", func() {
```
natalieparellano, line 130: Should this be changed to say "defaults to linux"?

testhelpers/testhelpers.go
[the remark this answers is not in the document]
micahyoung: Yeah, that totally makes sense. I'll make another attempt to clean up the outstanding tests with real base image needs. It does seem now there aren't any that need a "runnable" image if we remove the `CopyFromContainer` and use the diff ids instead. Then we'd land on just a couple tests (e.g. `Rebase`) and couple layer ones using a  `FromBaseImage(minimalBaseImage())` with the magic base image created on the fly, then we don't care where the base layers come from.

remote/remote_test.go
ekcasey: Agree, but I am going to merge to avoid delaying this large change any longer than necessary. We can also PR more test cleanup.

ekcasey:
Thanks for all the hard work @micahyoung :)

PR #43, opened 2020-04-22
Add support for writing Windows layer tars

micahyoung:
- Purpose: Windows daemon implementations require a more strict tar format than Linux layers. Windows base layers are proprietary and not authored by Docker or a normal user. However, non-base layers for either OS are typically created by `RUN`, `COPY`, etc in the Docker daemon and can now be created by the functionality introduced here. The WindowsWriter interface follows the pattern of a tar.Writer that is already often used for Linux layers, but with added internal state management for Windows needs. We also introduce an interface for TarWriter and a factory that will return the proper writer depending on an image (with minimal detection logic).
- A Linux, non-base layer for a container file `/myparent/myfile.txt` only requires the following:
  - `/myparent/mystuff.txt`      (required file entry; note that it implicitly defines a parent directory and that it is an absolute path)
- A Windows, non-base layer for a container file `c:\myparent\myfile.txt` requires the following:
  - `Files`                      (directory entry; note that it is a relative path)
  - `Files/myparent`             (directory entry)
  - `Files/myparent/myfile.txt`  (file entry)
  - `Hives`                      (directory entry; can be followed by child Windows reg hive deltas file entries though not required)

Signed-off-by: Andrew Meyer <meyeran@vmware.com>


micahyoung:
Here's [an example](https://github.com/buildpacks/pack/blob/c9846d205633499cb7c33a6f11bbdefec09c3e0c/internal/archive/archive.go#L45-L50) of the factory and writer in our draft PR for `pack`.

layer/windows_writer_test.go
```diff
@@ -0,0 +1,229 @@
...
+				h.AssertEq(t, th.Typeflag, byte(tar.TypeReg))
+
+				_, err = tr.Next()
+				h.AssertError(t, err, "EOF")
+			})
+		})
+	})
+
+	when("#Close", func() {
+		it("writes required parent dirs on empty image", func() {
```
ameyer-pivotal, line 202: Typo: should say `empty layer`

layer/windows_writer_test.go
```diff
@@ -0,0 +1,229 @@
...
+				th, _ = tr.Next()
+				h.AssertEq(t, th.Name, "Files/cnb/my-file")
+				h.AssertEq(t, th.Typeflag, byte(tar.TypeReg))
+
+				_, err = tr.Next()
+				h.AssertError(t, err, "EOF")
+			})
+		})
+
+		when("header path begins with dot", func() {
```
ekcasey, line 159: I am not sure relative path inputs even make sense in the context of a layer. We could require file paths to be absolute, or if this just happens to work w/o extra effort, maybe just remove the test so as not to imply that this is a recommended usage.

layer/windows_writer.go
```diff
@@ -0,0 +1,95 @@
...
+		tarWriter:          tar.NewWriter(fileWriter),
+		writtenParentPaths: map[string]bool{},
+	}
+}
+
+func (w *WindowsWriter) Write(content []byte) (int, error) {
+	return w.tarWriter.Write(content)
+}
+
+func (w *WindowsWriter) WriteHeader(header *tar.Header) error {
```
ekcasey, line 26: The behavior of `WriteHeader` when using the `WindowsWriter` is going to be different than that of the linux `tar.Writer` because this implementation writes **all** parent dir headers and `tar.Writer` does not. I think we either want a linux implementation that would, for example write `/cnb` if this was called with `/cnb/lifecycle` or we only want this implementation to only add the special dirs required by windows (I lean towards the latter so that users have control over the attributes of all normal files).

layer/tar_writer_factory.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/buildpacks/imgutil"
+	"github.com/buildpacks/imgutil/archive"
+)
+
+type tarWriterFactory struct {
+	os string
+}
+
+func NewTarWriterFactory(image imgutil.Image) (archive.TarWriterFactory, error) {
```
ekcasey, line 15: I think we should return the concrete struct here so we can return structs and accept interfaces when we don't have a compelling reason not to.

We can export `layer.TarWriterFactory` and it's field.

archive/tar_writer_factory.go
```diff
@@ -0,0 +1,24 @@
...
+	"io"
+)
+
+type TarWriter interface {
+	WriteHeader(hdr *tar.Header) error
+	Write(b []byte) (int, error)
+	Close() error
+}
+
+type TarWriterFactory interface {
```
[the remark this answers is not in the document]
ameyer-pivotal: Agreed. We can remove it

archive/tar_writer_factory.go
```diff
@@ -0,0 +1,24 @@
...
+	NewTarWriter(io.Writer) TarWriter
+}
+
+var DefaultTarWriterFactory = defaultTarWriterFactory{}
+
+type defaultTarWriterFactory struct{}
+
+func (defaultTarWriterFactory) NewTarWriter(w io.Writer) TarWriter {
+	return tar.NewWriter(w)
+}
```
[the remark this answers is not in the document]
ameyer-pivotal: It's a simple definition, and we'll also be removing the factory interface, so perhaps that's more reason to remove this and just define it where needed. Will look into it.

layer/windows_writer.go
```diff
@@ -0,0 +1,95 @@
...
+		return w.writeDirHeader(header)
+	}
+	return w.tarWriter.WriteHeader(header)
+}
+
+func (w *WindowsWriter) Close() error {
+	if err := w.initializeLayer(); err != nil {
+		return err
+	}
+	return w.tarWriter.Close()
```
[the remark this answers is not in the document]
ameyer-pivotal: Will do.

layer/windows_writer_test.go
ameyer-pivotal: That's a good point. I think we were trying too hard to make it behave like a normal tar writer. But you're right in that this is purely for writing layers, which require absolute paths.

layer/windows_writer.go
ameyer-pivotal: I'll argue a bit that in the windows case all parent dirs are "special" because they're required for creating a valid layer tar. If we think of "writing everything necessary for a layer" as the feature that we want parity on between the different types of writers, then we can say that the two writers behave equivalently.

I would also lean toward the latter if I didn't see every parent dir as required/special, so my vote would be to leave as is. Thoughts?

layer/tar_writer_factory.go
ameyer-pivotal: Good point. Will look into

layer/windows_writer.go
ameyer-pivotal: An alternative that we thought of was to create an interface that is more about writing layers, then create two concrete implementations, one for each OS. While I think that could be good to move towards in the long run, we opted not to do it right now to minimize the churn that would be required in dependent repos.

archive/tar_writer_factory.go
```diff
@@ -0,0 +1,24 @@
+package archive
+
+import (
+	"archive/tar"
+	"io"
+)
+
+type TarWriter interface {
```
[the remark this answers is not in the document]
ameyer-pivotal: We could rename it, but I worry about the implications it would have, though mostly aesthetic. Pack uses this interface in tar-writing code that is not necessarily related to layers. Now maybe Pack could have its own interface that mitigates this, but I'll need to dig more to see if it would compile, given the interface for the factory calls for the use of this interface (however, if that factory interface goes away and is defined by the consumer, we might be able to do this). I'll just try it and see where I land.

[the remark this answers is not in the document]
ameyer-pivotal: Resolved. Moving out of this library altogether, along with the factories.

PR #46, opened 2020-05-27
Gracefully handle any 404 status code when initializing images

matthewmcnew:
- Harbor 2 appears to return unexpected Code "NOT_FOUND" with a 404
- https://github.com/buildpacks/lifecycle/issues/303

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -92,8 +92,8 @@ func newV1Image(keychain authn.Keychain, repoName string) (v1.Image, error) {
 	image, err := remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport))
 	if err != nil {
 		if transportErr, ok := err.(*transport.Error); ok && len(transportErr.Errors) > 0 {
-			switch transportErr.Errors[0].Code {
-			case transport.UnauthorizedErrorCode, transport.ManifestUnknownErrorCode:
+			switch transportErr.StatusCode {
+			case http.StatusNotFound, http.StatusUnauthorized:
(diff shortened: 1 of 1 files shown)


remote/remote.go
```diff
@@ -92,8 +92,8 @@ func newV1Image(keychain authn.Keychain, repoName string) (v1.Image, error) {
 	image, err := remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport))
 	if err != nil {
 		if transportErr, ok := err.(*transport.Error); ok && len(transportErr.Errors) > 0 {
-			switch transportErr.Errors[0].Code {
-			case transport.UnauthorizedErrorCode, transport.ManifestUnknownErrorCode:
+			switch transportErr.StatusCode {
```
ekcasey, line 95: I think adding `tansport.NameUnknown` to the case here may fix our issue while still working with the domain specific error codes (unless Harbor doesn't comply with the registry spec)
*https://github.com/docker/distribution/blob/master/docs/spec/api.md#manifest

matthewmcnew: I believe we saw Harbor return a "NOT_FOUND" which would not be handled by NameUnknown.
```
{"errors":[{"code":"NOT_FOUND","message":"artifact demo/spring-music:v2 not found"}]}
```

ekcasey: FWIW, I think harbor is wrong here ;) but I'll concede that checking the status code is safer

matthewmcnew: Silly Harbor

PR #47, opened 2020-06-09
Improve Windows Header defaults

micahyoung:
The two commits here attempt to generate more valid Windows headers for layer entries.

The first commit validates that incoming header paths are valid POSIX format and not accidental windows paths. No validation happens in `tar.WriteHeader` and this mostly helps us find incorrect usages of `filepath` in consumers of `imgutil`.

The second commit sets basic default PAX record permissions for entries, based on the incoming header's `Uid` and `Gid`. This is needed for images to run on k8s (ContainerD) containers. ContainerD accepts layers with entries missing PAX records but in the generated containers, those files have no permissions and are inaccessible. However, Docker instead generates default permissions for those same entries and in the containers the files are owned by `USER`. So this implementation roughly attempts to do what Docker does for all entries, albeit based on `Uid` and `Gid` to the extent possible.

They map as follows:
- Both UID and GID are 0 => Set Owner + Primary Group to BUILTIN\Administrators
- Any other values => Set Owner + Primary Group to BUILTIN\Users (default group for any real user on the system)

This is a basic implementaion to start with, but a more fine-grained mapping will likely be preferable, perhaps informed by how Linux NTFS or SMB drivers map permissions.

### DIFF
--- a/layer/windows_writer.go
+++ b/layer/windows_writer.go
@@ -39,10 +39,23 @@ func (w *WindowsWriter) WriteHeader(header *tar.Header) error {
 		return err
 	}
 
+	if header.PAXRecords == nil {
+		header.Format = tar.FormatPAX
+
+		header.PAXRecords = initializePaxPermissions(header)
+	}
+
 	if header.Typeflag == tar.TypeDir {
 		return w.writeDirHeader(header)
 	}
 	return w.tarWriter.WriteHeader(header)
+}
+
+func initializePaxPermissions(header *tar.Header) map[string]string {
+	if header.Uid == 0 && header.Gid == 0 {
+		return map[string]string{"MSWINDOWS.rawsd": AdministratratorOwnerAndGroupSID}
+	}
+	return map[string]string{"MSWINDOWS.rawsd": UserOwnerAndGroupSID}
 }
 
 func (w *WindowsWriter) Close() (err error) {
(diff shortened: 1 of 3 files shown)


micahyoung:
@ameyer-pivotal Would you mind taking a look at this? Assuming it looks good to you, we can assign more folks to review.

micahyoung:
Holding off on this one for now - we found some issues in Docker

micahyoung:
After quite a bit of experimentation, I couldn't get the suspected error scenario to reproduce so I'm reopening as-is.

This has now worked quite consistently for every test scenario I could come up with and I believe the suspected error I was seeing before closing this issue may have actually been due to switching back and forth from Hyper-V to process isolation for different phases of `local` image creation. Hyper-V isolation has proven to be quite buggy (and many times slower) with volumes and bind mounts, and I now I suspect image creation as well. In `pack` and `lifecycle` we're explicitly using process isolation wherever possible as it is faster and matches k8s and Docker swarm production defaults.

layer/windows_writer.go
```diff
@@ -94,9 +111,17 @@ func (w *WindowsWriter) writeDirHeader(header *tar.Header) error {
...
 	if err := w.tarWriter.WriteHeader(header); err != nil {
 		return err
 	}
 	w.writtenParentPaths[header.Name] = true
 	return nil
 }
+
+func (w *WindowsWriter) validateHeader(header *tar.Header) error {
+	if !path.IsAbs(header.Name) {
+		return fmt.Errorf("invalid header came: must be absolute, posix path: %s", header.Name)
```
ekcasey, line 124:
```suggestion
		return fmt.Errorf("invalid header name: must be absolute, posix path: %s", header.Name)
```

micahyoung:
I fixed the typo and rebased. Ready to go. Thanks @ekcasey !

PR #49, opened 2020-06-18
Add the ability to create registry with authentication

natalieparellano:
This PR is intended to add functionality to the docker registry in `testhelpers` so that this package can be used in the lifecycle acceptance tests. We need the ability to create a registry with authentication.

I am not sure about all of the `registry.username != ""` in `registry.Start()` - maybe it's better to have a different method or something.

Signed-off-by: Natalie Arellano <narellano@vmware.com>

### DIFF
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"testing"
 	"time"
@@ -47,7 +46,7 @@ func NewDockerRegistryWithAuth(dockerConfigDir string) *DockerRegistry {
 }
 
 func (r *DockerRegistry) Start(t *testing.T) {
-	t.Log("run r")
+	t.Log("run registry")
 	t.Helper()
 
 	ctx := context.Background()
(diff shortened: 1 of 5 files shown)


testhelpers/docker_registry.go
```diff
@@ -68,3 +123,52 @@ func (registry *DockerRegistry) Stop(t *testing.T) {
 		DockerCli(t).ContainerRemove(context.TODO(), registry.Name, types.ContainerRemoveOptions{Force: true})
 	}
 }
+
+func (registry *DockerRegistry) RepoName(name string) string {
```
ekcasey, line 127:
```suggestion
func (r *DockerRegistry) RepoName(name string) string {
```
Not sure why this and others had `registry` as a receiver instead of `r` but we should probably fix it while we are looking at this.

testhelpers/testhelpers.go
```diff
@@ -184,12 +184,26 @@ func PushImage(dockerCli dockercli.CommonAPIClient, ref string) error {
...
-		return "", err
+		return "", errors.Wrap(err, "making new request")
+	}
+
+	if len(headers) > 0 {
+		for key, val := range headers[0] {
+			request.Header.Set(key, val)
+		}
+	}
+
```
ekcasey, line 200:
```suggestion
	for key, val := range headers {
		request.Header.Set(key, val)
	}

```

testhelpers/testhelpers.go
```diff
@@ -184,12 +184,26 @@ func PushImage(dockerCli dockercli.CommonAPIClient, ref string) error {
...
+	resp, err := client.Do(request)
```
ekcasey, line 201:
```suggestion
	resp, err := http.DefaultClient.Do(request)
```
If this is the only place we use the client why not inline like this and remove the `client` var?

testhelpers/testhelpers.go
natalieparellano: This won't compile because `headers` is a list of maps. I did it this way (variadic headers) to avoid breaking current consumers of `HTTPGetE` - but it looks like the only usage within the repo is within docker_registry.go. Should I make `headers` a required argument?

natalieparellano: About to push some code that has this change (making it required) - LMK if you want me to put it back :)

natalieparellano:
Ah, you know what - I missed the part about using `RepoName` in remote_test.go - let me check again.

natalieparellano:
Should be good now on `RepoName`...

testhelpers/testhelpers.go
ekcasey: The changes look good. Sorry I assumed, rather than checking the type of `headers`. In any case in seems wrong to except variadic headers and then only respect the first one. Making it required is better imo.

testhelpers/docker_registry.go
```diff
@@ -22,13 +32,23 @@ var registryImageNames = map[string]string{
...
+		username:        RandString(10),
+		password:        RandString(10),
+		DockerDirectory: dockerConfigDir,
+	}
 }
 
-func (registry *DockerRegistry) Start(t *testing.T) {
-	t.Log("run registry")
+func (r *DockerRegistry) Start(t *testing.T) {
+	t.Log("run r")
```
ekcasey, line 50:
```suggestion
	t.Log("run registry")
```
Search and replace? :)

natalieparellano: Good catch!

natalieparellano:
> It seems like windows tests are now failing b/c the the `tar` command used to generate `htpasswd` doesn't work on windows. We can probably copy whatever `pack` does to solve that problem (programmatically generate a tar, I am assuming?)

Oh saddd... yes that is what pack is doing. It is using a bunch of archive utilities that didn't exist in lifecycle, hence I was lazy. I will look into it.

natalieparellano:
Windows is fixed now :)

PR #53, opened 2020-08-04
Fix remote rebase losing platform values

mvalliath:
**Problem:**
When `pack rebase` is run on a windows container image, the resulting container image is unable to run on a Windows daemon: the problem is that the image loses its **os.Platform** metadata (for both platforms) during the process. We’ve raised an [issue with google/go-containerregistry](https://github.com/google/go-containerregistry/issues/751), but have implemented a stop-gap within this PR.

**Related:**
* The following PR is intended to address https://github.com/buildpacks/pack/issues/770. For `pack` to have this resolved, it should be rebuilt with code from this PR.
* https://github.com/google/go-containerregistry/issues/751

Signed-off-by: Anthony Emengo <aemengo@vmware.com>

@aemengo @micahyoung

### DIFF
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -667,6 +667,10 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 				afterLayer4DiffID := afterInspect.RootFS.Layers[len(afterInspect.RootFS.Layers)-1]
 				h.AssertEq(t, imgLayer2DiffID, afterLayer4DiffID)
+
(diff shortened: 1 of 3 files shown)


ekcasey:
> The following PR is intended to address buildpacks/pack#770. For pack to have this resolved, it should be rebuilt with code from this PR.

I think `pack` uses `Rebase` from the lifecycle (as do other platform). We will need to update the lifecycle to pull in the newest `imgutil`.

PR #56, opened 2020-08-14
Don't duplicate environment variables

natalieparellano:
On Windows, ignore case when determining if variable is a duplicate.
Use the casing of the key that was passed in.

Signed-off-by: Yael Harel <yharel@vmware.com>

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -235,6 +235,20 @@ func (i *Image) SetLabel(key, val string) error {
 }
 
 func (i *Image) SetEnv(key, val string) error {
+	ignoreCase := i.inspect.Os == "windows"
+	for idx, kv := range i.inspect.Config.Env {
+		parts := strings.SplitN(kv, "=", 2)
+		foundKey := parts[0]
+		searchKey := key
+		if ignoreCase {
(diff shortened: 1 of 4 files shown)


natalieparellano:
Fixes https://github.com/buildpacks/lifecycle/issues/385

natalieparellano:
@micahyoung @ameyer-pivotal is it possible you could give this a glance from a Windows perspective?

natalieparellano:
We are planning to add a `ListEnv()` function as well which would allow us to unit-test the exporter. We will update the PR after lunch.

natalieparellano:
Upon further reflection, `ListEnv()` isn't going to help us, because the exporter tests are currently using a fake image with its own implementation of `imgutil.Image`. We don't have acceptance tests for the exporter yet, so we'll manually test this.

natalieparellano:
We confirmed that this fixes buildpacks/lifecycle#385

nebhale:
@micahyoung I'd be super surprised if anyone but us is creating images with duplicate env vars at this point.  The failure mode is so quick and so completely, that something like JIB would encounter it once during development (like us), solve it (like us), and never generate an image like that again.

PR #57, opened 2020-08-18
add methods to differentiate windows and linux test images

dwillist:
Signed-off-by: dwillist <dthornton@vmware.com>


fakes/image.go
```diff
@@ -35,6 +35,27 @@ func NewImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
...
+		createdAt:     time.Now(),
+		savedNames:    map[string]bool{},
+		os:            "windows",
+		osVersion:     "",
+		architecture:  "amd64",
+	}
+}
+
+func NewLinuxImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
+	return NewImage(name, topLayerSha, identifier)
```
yaelharel, line 57: I guess we're planning to delete `NewImage` at some point. Did we create an issue for that? Maybe it's worth mentioning it in this issue.

ekcasey: seems like we might want an unexported `newImage` that sets everything besides `os`. `NewWindowsImage` and `NewLinuxImage` can call that helper and then set `os`. I worry about drift/bugs if the shared setup is duplicated in `NewLinuxImage` and `NewWindowsImage`. 

Another option: we could refactor NewImage to take option functions that would include a `WithOS` option. That might be best. We currently have many tests that pass `""` and `nil` params into `fakes.NewImage`. Options functions would allow us to set only the relevant fields for each test and default everything else.

yaelharel: @ekcasey, your ideas are great. I thought that we need to keep `NewImage` since some old interface is use it as is.
If I'm wrong, I totally agree with you! I think that I prefer your first option but I'm also fine with the second one.

ekcasey: > I thought that we need to keep NewImage since some old interface is use it as is

If we get rid of `NewImage` or change its interface, we will have to update all the tests when we bump to the latest `imgutil` in either `lifecycle ` or `pack`. However, we should probably bite the bullet and do it if the new interface is the right choice in the long run.

yaelharel: Sounds good! I think it's worth doing it!

fakes/image.go
```diff
@@ -35,6 +35,27 @@ func NewImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
 		architecture:  "amd64",
 	}
 }
+func NewWindowsImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
```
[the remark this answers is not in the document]
dwillist: Pulling the OS up to the top level was done strictly for visibility. Reasoning about what windows container backed is being used and tested in some of [these](https://github.com/buildpacks/pack/pull/809) tests was difficult & likely the reason many were omitted.

fakes/image.go
dwillist: Ya that sound much better than what I have here. 👍

jromero: > This ^ is probably another good reason to migrate all configuration to option functions or setters (and away from positional params). That way it is easy to extend or change in the future without breaking compilation for every test.

🖐️ for option functions over setters. Mutability is always a good source of bugs.

dwillist:
Going to close this. I've moved most of this logic out into pack itself. Instead will just do the refactor mentioned above.

PR #60, opened 2020-08-27
Add #Labels to imgutil.Image

jabrown85:
We've got [an upcoming feature](https://github.com/buildpacks/lifecycle/issues/390) that requires getting all the lables on an image instead of a specificly named label.

### DIFF
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -20,7 +20,7 @@ import (
 
 func NewImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
 	return &Image{
-		labels:        map[string]string{},
+		labels:        nil,
 		env:           map[string]string{},
 		topLayerSha:   topLayerSha,
 		identifier:    identifier,
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -67,6 +67,10 @@ func (i *Image) Label(key string) (string, error) {
 	return i.labels[key], nil
 }
 
+func (i *Image) Labels() (map[string]string, error) {
+	return i.labels, nil
+}
+
 func (i *Image) OS() (string, error) {
 	return i.os, nil
 }
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -101,6 +101,9 @@ func (i *Image) Rebase(baseTopLayer string, newBase imgutil.Image) error {
 }
 
 func (i *Image) SetLabel(k string, v string) error {
+	if i.labels == nil {
+		i.labels = map[string]string{}
+	}
 	i.labels[k] = v
 	return nil
 }
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -230,7 +230,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				h.AssertNil(t, h.DockerRmi(dockerClient, repoName))
 			})
 
-			it("returns an empty map", func() {
+			it("returns nil", func() {
 				img, err := local.NewImage(repoName, dockerClient, local.FromBaseImage(repoName))
 				h.AssertNil(t, err)
 
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -190,7 +190,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 	})
 
 	when("#Labels", func() {
-		when("image exists", func() {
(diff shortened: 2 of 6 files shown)


local/local_test.go
```diff
@@ -189,6 +189,35 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 
+	when("#Labels", func() {
+		when("image exists", func() {
```
ekcasey, line 193: Do we want a test case for when the image does not exist? Or when there are no labels? I think right now both of these scenarios return `nil`. This opens the caller up to more potential errors than returning an empty map would, but is probably accurate  given that the image config json has `nil` rather than `{}`, `labels` in these cases.

Instead of changing `img.Labels()` to `return map[string]string` when the the config file has `nil` labels I think we should change our empty image such that labels are never `nil`.

remote/remote_test.go
```diff
@@ -196,6 +196,31 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 
+	when("#Labels", func() {
+		when("image exists", func() {
```
ekcasey, line 200: Same comment as above regarding cases where the image does not exist or no labels have been set.

fakes/image.go
```diff
@@ -67,6 +67,10 @@ func (i *Image) Label(key string) (string, error) {
 	return i.labels[key], nil
 }
 
+func (i *Image) Labels() (map[string]string, error) {
+	return i.labels, nil
```
ekcasey, line 71: The fake image never has `nil` labels, and therefore code that passes tests against the fake image might be vulernable to nil pointer exceptions when provided a local or remote image at runtime. We should either change the local and remote images to return an empty map or change the fake image to return `nil`.

local/local_test.go
jabrown85: I added test for missing image and when there are no labels on an existing image.

Isn't `nil` eq to an empty map in this case? The tests I wrote seem to imply there is no functional difference between the two. 

I know `nil` or empty maps are not writeable, is that the concern? Do you think we need to `make(map[string]string)`?

fakes/image.go
jabrown85: See previous response, I don't think we are in danger of a pointer exception here. `nil` behaves like an empty map already.

remote/remote_test.go
jabrown85: See previous response, I don't think we are in danger of a pointer exception here. `nil` behaves like an empty map already.

local/local_test.go
ekcasey: Yeah you are right that the circumstances where this would become an issue are a bit esoteric. `SetLabel` works without errors because we initialize an empty map before adding a key if the labels are `nil`. This would only become an issue if you you called `img.Label()` and then tried set a key on the returned value (rather than calling `SetLabel` on the image).

I am fine leaving local/remote as is but let's change the fake to match the implementations just to be cautious.

jabrown85: Alright, I pushed up a change to `SetLabel` for the fake implementation. I think that is what you meant for me to do..right?

ekcasey: Yes we need that change in `SetLabel`. Also, I think we should change `fakes.NewImage` to initialize the image with `nil` labels instead of empty labels https://github.com/buildpacks/imgutil/blob/main/fakes/image.go#L23

Then the behavior of `Labels` will be the same for fake, local, and remote images in cases where there are no labels.

jabrown85: Ahh, ok. I see.

local/local_test.go
```diff
@@ -189,6 +189,69 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertNil(t, err)
+
+				labels, err := img.Labels()
+				h.AssertNil(t, err)
+				h.AssertEq(t, 0, len(labels))
+			})
+		})
+
+		when("image NOT exists", func() {
+			it("returns an empty map", func() {
```
ekcasey, line 244: Not to be too pedantic here but I think it is returning `nil` not an empty map right?

jabrown85: Yup. I updated the names, I think we are good.

PR #64, opened 2020-10-08
Adds support for creating Windows scratch-equivalent baselayers

micahyoung:
This allows `FROM scratch`-style Windows images (and buildpackages) to be stored on WCOW daemons (required for https://github.com/buildpacks/pack/pull/840).

This is needed because `imgutil` in `local` scenarios requires all images to be stored on a Docker daemon. This works fine for Linux since the Docker daemon has pretty loose validation for Linux images (only requires the correct image metadata). But WCOW daemons, on the other hand, are most strict and require a specific baselayer as well. This difference is likely because there is no current `FROM scratch` use-case for Windows since it's not currently possible to statically compile a Windows binary with enough libraries to run in an empty container.

The smallest Microsoft-authored baselayer that can be stored on a daemon is from nanoserver, which is 100MB, has license restrictions for use, is a foreign layer that requires public internet access to pull, changes frequently due to patches, is coupled to the OS version (`nanoserver:2004` images can't be pulled by Windows 1809), and currently has no long-term-supported version. For these reasons, it is unsuitable as a scratch-equivalent baselayer for something like buildpackages.

This implementation was derived from the files required by the [moby](https://github.com/moby/moby/blob/master/daemon/graphdriver/windows/windows.go) and [hcsshim](https://github.com/microsoft/hcsshim/blob/master/internal/wclayer/legacy.go) codebases. It creates a layer with the 6 files necessary so an image that uses it can be stored on a daemon. There is no documented API contract for these files unfortunately which is a big downside. However the benefits in contrast to using nanoserver are that it can be generated on the fly on any OS, is created with OSS tools, has no distribution limitations, works with all current daemon versions, is only a few bytes, is trivial enough to hopefully not require regular patches, and if it does need to be patched, can be replaced with `Image.Rebase()`. The images with this layer can also be pulled and inspected with daemon-based tools like `dive` and presumably anything using the Docker API.

This implementation includes a new generator to create the `imgutil` source file with the encoded content of the most complex file - the  baselayer's BCD hive file. This is needed only because the [hivex](https://github.com/libguestfs/hivex) C++ shared library is difficult to compile for OSes besides Linux and would introduce a surprising runtime requirement for consumers of `imgutil` on any OS.

I manually tested the layer against the main supported versions of Docker Desktop for Windows (1809, 2004) and Docker for Windows Server (1809, 2004). Additionally, a nearly identical layer has been used as an `imgutil` [test fixture](https://github.com/buildpacks/imgutil/blob/cea9fc5483726ce3864109c335cc1bf7c19a42ae/testhelpers/testhelpers.go#L297) for the past several months without issue.


micahyoung:
After discussion in Slack, I realized I overemphasized the importance of this scratch base layer for `imgutil` scenarios. But in fact, any Windows OCI image must have some base layer like this one to allow `docker pull`, `docker load`, ggcr `daemon.Write()` or anything else that writes and image to a WCOW daemon, even if the base layer is not runnable.

A user who is not interested in using `imgutil.Image`s could still benefit from using `layer.WindowsBaseLayer` with something like ggcr `mutate.AppendLayer(empty.Image, tarball.LayerFromReader(layer.WindowsBaseLayer))`

ameyer-pivotal (approved):

ekcasey (changes requested):
@micahyoung Thanks for all of your hard work here. I appreciate the effort that went into creating the nice `go generate` interface.

I left some comments the mostly boil down to:
1. some very minor cleanup in the testhelpers
2. can we make the Makefile target smart enough to avoid slow generation steps when possible? If not maybe we should pull it out of the `test` target it leave it to users who edit the bsdgen code to run manually when appropriate?

Makefile
```diff
@@ -20,5 +21,22 @@ lint: install-golangci-lint
...
+else
+	@echo "> Cannot generate on Windows"
+endif
+
+layer/bcdhive_generated.go: layer/windows_baselayer.go tools/bcdhive_generator/.build
+ifneq ($(OS),Windows_NT)
+	$(GOCMD) generate ./...
+else
+	@echo "> Cannot generate on Windows"
+endif
```
micahyoung, line 39: @ekcasey I changed targets to be the actual files, with the `docker build` target generating a gitignored `.build` file that is timestamp-checked along with the rest of the contents.

This still feels a little busy... I wonder if putting all this into either sub-Makefiles would be preferrable. Otherwise, we could also just fold all this into a single target.

Makefile
```diff
@@ -1,5 +1,6 @@
 # Go parameters
 GOCMD?=go
+GO_VERSION=$(subst go,,$(word 3, $(shell go version)))  # parses "go version go1.15 darwin/amd64"
```
micahyoung, line 3: I figured the current go version might be safest to key off

testhelpers/testhelpers.go
```diff
@@ -222,137 +220,79 @@ func CreateSingleFileTarReader(path, txt string) io.ReadCloser {
...
+		}
 
-		err = writeTarSingleFileLinux(tw, path, txt) // Use the Linux writer, as this isn't a layer tar.
+		if err := tw.Close(); err != nil {
+			pw.CloseWithError(err)
+		}
+
+		if err := pw.Close(); err != nil {
+			pw.CloseWithError(err)
+		}
```
micahyoung, line 240: Closing the pipe writer was still needed, otherwise the reader would block.

Makefile
ekcasey: I think it's better to hardcode so that different users don't get different results based on the state of their machines.

If we want to be clever and avoid hardcoding in two places we could grab it from the [mod file](https://github.com/buildpacks/imgutil/blob/main/go.mod#L16) with `go list -m -f "{{.GoVersion}}"`.

testhelpers/testhelpers.go
```diff
@@ -272,21 +273,24 @@ func CreateSingleFileLayerTar(layerPath, txt, osType string) (string, error) {
...
 
 	baseLayer, err := layer.WindowsBaseLayer()
 	AssertNil(t, err)
 
-	baseLayerBytes, err := ioutil.ReadAll(baseLayer)
+	_, err = io.Copy(tarFile, baseLayer)
 	AssertNil(t, err)
 
-	AssertNil(t, ioutil.WriteFile(tarFile.Name(), baseLayerBytes, 0666))
+	AssertNil(t, tarFile.Close())
```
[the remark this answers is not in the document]
micahyoung: Ugh, good catch. Fixing now.

Makefile
micahyoung: Thanks for that command. I was struggling to figure out how to do exactly that!

Makefile
ekcasey: lgtm for now. feel free to submit a cleanup PR that combines them if you think it looks better

ekcasey (approved):

Issue #54, opened 2020-08-05
remote.NewImage() with remote.FromBaseImage() fails on Windows when provided a manifest list


natalieparellano:
E.g., 

`remote.NewImage(tags[0], authn.DefaultKeychain, remote.FromBaseImage("mcr.microsoft.com/windows/nanoserver:1809")` (a manifest list)

fails with 

`connect to repo store 'mcr.microsoft.com/windows/nanoserver:1809': no child with platform amd64/linux in index mcr.microsoft.com/windows/nanoserver:1809`

Whereas

`remote.NewImage(tags[0], authn.DefaultKeychain, remote.FromBaseImage("mcr.microsoft.com/windows/nanoserver:1809-amd64")` (an image index)

works just fine.

ekcasey:
Makes sense. I think remote.NewImage, needs a `WithPlatform` option that we can pass through to `ggcr` so we can select the correct manifest from the list.

jjbustamante:
Hi @ekcasey @natalieparellano, I am planning to give it a shot to this issue, right now I am receiving help to create a VM with windows where I can actually have administration permissions to install tools on it, once that's done, I will prepare my environment and try to reproduce the issue.

PR #66, opened 2020-10-12
Adds setters for OS/OSVersion/Architecture

micahyoung:
Required to create valid, empty images without `FromBaseImage`

This will be important for Windows buildpackages in pack (currently draft PR: https://github.com/buildpacks/pack/pull/840), which will not use `FromBaseImage`

### DIFF
--- a/image.go
+++ b/image.go
@@ -38,6 +38,9 @@ type Image interface {
 	SetEntrypoint(...string) error
 	SetWorkingDir(string) error
 	SetCmd(...string) error
+	SetOS(string) error
+	SetOSVersion(string) error
+	SetArchitecture(string) error
 	Rebase(string, Image) error
 	AddLayer(path string) error
 	AddLayerWithDiffID(path, diffID string) error
--- a/local/local.go
+++ b/local/local.go
@@ -242,6 +242,21 @@ func (i *Image) SetLabel(key, val string) error {
 	return nil
 }
 
+func (i *Image) SetOS(osVal string) error {
+	i.inspect.Os = osVal
+	return nil
+}
+
+func (i *Image) SetOSVersion(osVersion string) error {
+	i.inspect.OsVersion = osVersion
+	return nil
+}
+
+func (i *Image) SetArchitecture(architecture string) error {
+	i.inspect.Architecture = architecture
+	return nil
+}
+
 func (i *Image) RemoveLabel(key string) error {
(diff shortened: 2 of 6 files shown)


ekcasey (approved):

image.go
```diff
@@ -38,6 +38,9 @@ type Image interface {
 	SetEntrypoint(...string) error
 	SetWorkingDir(string) error
 	SetCmd(...string) error
+	SetOS(string) error
+	SetOSVersion(string) error
+	SetArchitecture(string) error
```
ekcasey, line 43: Eventually I feel like these options make more sense as options on `NewImage`. Having constructor options instead of a setter would convey that is not valid to change these values for any image other than an empty image. But I can see how the awkwardness of our existing `FromBase` option makes that difficult to support without broader changes.

ekcasey (approved):

ameyer-pivotal (approved):
I can see @ekcasey's point about setting the OS to something other than the daemon OS. Otherwise 👍

image.go
micahyoung: Yeah, good call. I was also wondering how it will intersect with the needs of https://github.com/buildpacks/imgutil/issues/54 which would also need very similar looking flags but to instead filter the `FromBaseImage` platform.

local/local.go
```diff
@@ -242,6 +242,23 @@ func (i *Image) SetLabel(key, val string) error {
 	return nil
 }
 
+func (i *Image) SetOS(osVal string) error {
+	if i.inspect.Os != osVal {
```
micahyoung, line 246: I changed this to not set anything, and instead compare against the current value of `i.inspect.Os` which is already the daemon's default from `defaultInspect`. If we ever get a multi-OS daemon, we'll have to revisit of course.

local/local.go
```diff
@@ -242,6 +242,23 @@ func (i *Image) SetLabel(key, val string) error {
 	return nil
 }
 
+func (i *Image) SetOS(osVal string) error {
+	if osVal != i.inspect.Os {
+		return fmt.Errorf(`invalid os: must match the daemon: "%s"`, i.inspect.Os)
+	}
+	return nil
```
micahyoung, line 249: I changed this to not set anything, and instead compare against the current value of `i.inspect.Os` which is already the daemon's default from `defaultInspect`. If we ever get a multi-OS daemon, we'll have to revisit of course.

PR #68, opened 2020-10-17
feat: use released version of Docker

greut:
this project is using a non-released version of Docker, hence it's unlikely that anyone is actually running the software it's built against. This patches pull docker v19.03.13 instead.

Sadly this swarmkit feature is not part of it yet.

https://github.com/moby/moby/commit/d363a1881ec256e1be5a48a90046ff16e84ddc02


local/local.go
```diff
@@ -685,7 +685,6 @@ func defaultInspect(docker client.CommonAPIClient) (types.ImageInspect, error) {
 
 	return types.ImageInspect{
 		Os:           daemonInfo.OSType,
-		OsVersion:    daemonInfo.OSVersion,
```
micahyoung, line 688: I think removing the default value for `OsVersion` is fine - and improvement actually. A blank value will generate more consistent images that any daemon can pull. The existing behavior is that a new `local` image - using the default `OsVersion` from daemon - could only be pulled by other daemons of equal or greater version.

micahyoung (approved):

ekcasey (approved):

PR #81, opened 2020-12-04
Add fallback behavior if performance optimizations fail

### LINKED ISSUE #63: local.NewImage fails because docker.SaveImage returns multiple manifests.
Periodically when executing the lifecycle against the daemon errors like this occur:
```
ERROR: rebase working image: manifest.json had unexpected number of entries: 2
```

This happens when we are saving an image from the docker daemon to disk and we unexpectedly get more than one manifest. https://github.com/buildpacks/imgutil/blob/afd98bd2f655b1d828f1acded8dd81429ca8b9b5/local/local.go#L542-L544

We should figure out how we get into this state and determine a mitigation.

### LINKED ISSUE #80: Fallback to slow behavior if writing to daemon fails with performance optimization
We currently omit base image layers when writing to the docker daemon in order to improve performance. However some container runtimes like Podman that implement the daemon API do not allow an image load with missing layers. Therefore if our first attempt at writing fails, we should fall back on the slow/complete behavior and provide all layer tars.

context: https://github.com/buildpacks/pack/issues/925

ekcasey:
Resolves #80 
Resolves #63

When saving a daemon image we now first attempt a performant save with base layers omitted and, if that fails, fetch the base layers by `docker save`ing the image, and try again. I have also removed more complicated logic that was intended to omit layers reused from the previous image whenever possible. With the addition of the launch cache in lifecycle, neither lifecycle nor pack currently take advantage of this optimization and it adds complexity.

While intended to resolve #80 this also resolves #63 as a side effect because we now save an image to disk by ID, avoiding situations where `manifest.json` might contain multiple entries. This previously happened because we were passing a tag to docker save without expanding an implicit `:latest` tag. While `docker inspect` will treat `my/image` as `my/image:latest` `docker save` will return all images with repository `my/image`. While we could save using the expanded tag, ID is the most explicit and also avoid any potential race conditions.

This PR is intended to provide fixes for immediate problems with as little disruption as possible. I intend to follow up in the next two weeks with a redesign of the imgutil interface that addresses some of the domain modeling problems and improves the usage.

I attempted to test this manually with `podman` but ran into issues where the result of `docker inspect` was missing fields. I believe this is an incompatibility between `podman` and the docker client library rather than an `imgutil` problem. I did temporarily comment out the performance hack and rerun tests to ensure the fallback isn't broken. More robust/automated testing of the fallback will be easier to introduce after the refactor mentioned above so I have omitted it for now.


local/local.go
```diff
@@ -310,16 +269,19 @@ func (i *Image) TopLayer() (string, error) {
...
+	for l := range i.inspect.RootFS.Layers {
+		if i.inspect.RootFS.Layers[l] != diffID {
+			continue
+		}
+		if i.layerPaths[l] == "" {
+			if err := i.downloadBaseLayersOnce(); err != nil {
+				return nil, err
+			}
+		}
+		return os.Open(i.layerPaths[l])
```
jromero, line 281: An edge case for sure but wouldn't it be better to check if `i.layerPaths[l]` is valid/non-empty after downloading the base layers? Otherwise, a very cryptic error may be thrown if that edge case is ever encountered.

micahyoung (changes requested):
Looks good to me, except I'm running into some temp file or test image leakage or something - each time I run the `local` tests on Windows, it seams about 1GB of data gets left behind. Still investigating but wanted to let you know in the meantime.

micahyoung (approved):
Looks good to me. Took me a while to get caught up on the issues and the impact of removing the optimization but this simplify things as well (also `local` tests on Windows speed up by about 30%!).

PR #83, opened 2020-12-11
Fix make parental dirs for symlinks on untar

matejvasek:
Signed-off-by: Matej Vasek <mvasek@redhat.com>

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -623,6 +623,13 @@ func untar(r io.Reader, dest string) error {
 			}
 			fh.Close()
 		case tar.TypeSymlink:
+			_, err := os.Stat(filepath.Dir(path))
+			if os.IsNotExist(err) {
+				if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
+					return err
+				}
+			}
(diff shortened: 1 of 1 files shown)


local/local.go
```diff
@@ -624,12 +629,26 @@ func untar(r io.Reader, dest string) error {
 			fh.Close()
 		case tar.TypeSymlink:
 			if err := os.Symlink(hdr.Linkname, path); err != nil {
```
matejvasek, line 631: I don't understand why `os.Symlink` returns err when target doesn't exists.
When I tried to create new golang project it seems to not care about that. Also `ln` commands happily creates links with invalid destination.

matejvasek: Ahh this might be happening because parent dir doesn't exist!

matejvasek: Yep the tar I was receiving (from `podman`) is missing directories for some reason.

matejvasek: The `untar` function does create parental dirs for regular files, but not for symlinks. This PR has to be reworked!

PR #84, opened 2020-12-18
Update google/go-containerregistry

mattmoor:
This pulls in the estargz support in google/go-containerregistry, which enables buildpack builders to produce estargz layers by simply setting `GGCR_EXPERIMENT_ESTARGZ=1`.

Related: https://github.com/opencontainers/image-spec/issues/815


mattmoor:
Talked to @ekcasey in slack, I am going to try and get a GGCR release cut, and let dependabot do its magic.

PR #86, opened 2021-01-12
Add a ManifestSize function to the Image interface

yaelharel:
The function returns the size of the image manifest.
If a manifest doesn't exist, it returns 0.

Signed-off-by: Yael Harel <yharel@vmware.com>

### DIFF
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -57,6 +57,7 @@ type Image struct {
 	layerDir      string
 	workingDir    string
 	savedNames    map[string]bool
+	manifestSize  int64
 }
 
 func (i *Image) CreatedAt() (time.Time, error) {
@@ -410,4 +411,12 @@ func (i *Image) SavedNames() []string {
 	}
 
 	return names
+}
+
+func (i *Image) SetManifestSize(size int64) {
+	i.manifestSize = size
+}
+
+func (i *Image) ManifestSize() (int64, error) {
+	return i.manifestSize, nil
 }
--- a/image.go
+++ b/image.go
@@ -59,6 +59,8 @@ type Image interface {
 	OS() (string, error)
 	OSVersion() (string, error)
 	Architecture() (string, error)
+	// ManifestSize returns the size of the manifest. If a manifest doesn't exist, it returns 0.
+	ManifestSize() (int64, error)
 }
 
 type Identifier fmt.Stringer
--- a/local/local.go
+++ b/local/local.go
@@ -482,6 +482,10 @@ func (i *Image) Delete() error {
 	return err
 }
 
+func (i *Image) ManifestSize() (int64, error) {
+	return 0, nil
+}
+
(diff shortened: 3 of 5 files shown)


fakes/image.go
```diff
@@ -411,3 +412,11 @@ func (i *Image) SavedNames() []string {
 
 	return names
 }
+
+func (i *Image) SetManifestSize(size int64) {
+	i.manifestSize = size
+}
```
natalieparellano, line 418: Why would this be needed?

yaelharel: To test it in `lifecycle`.
I'll make a PR there soon.

yaelharel: Please see https://github.com/buildpacks/lifecycle/pull/498

yaelharel: I'm going to mark this conversation as resolved. Please reopen it if it's still unclear.

local/local.go
```diff
@@ -482,6 +482,10 @@ func (i *Image) Delete() error {
 	return err
 }
 
+func (i *Image) ManifestSize() (int64, error) {
+	return 0, nil
```
micahyoung, line 486: I'm not sure if manifest size is important for `local` but it does look like a manifest is generated elsewhere in the file, which you could potentially extract to a shared function and do a `len` on the return: https://github.com/buildpacks/imgutil/blob/ae3e2fcd82781e7c6aba7bc25f3e111467e44e11/local/local.go#L432-L438

yaelharel: This is interesting.
According to https://github.com/buildpacks/spec/issues/169:
```This field will not apply to the -daemon case, b/c no manifest has been generated.```
Let me know if you still think we should change the `ManifestSize` implementation.

Comment (micahyoung) [reply local/local.go:486]: Yeah, that makes sense. I missed the RFC the first time around. Any `local` manifest size would be hypothetical anyways so `0` feels fine.

File: image.go [hunk 3]
```diff
@@ -59,6 +59,8 @@ type Image interface {
 	OS() (string, error)
 	OSVersion() (string, error)
 	Architecture() (string, error)
+	// Size returns the size of the manifest. If a manifest doesn't exist, it returns 0.
```
natalieparellano, line 62:
```suggestion
	// ManifestSize returns the size of the manifest. If a manifest doesn't exist, it returns 0.
```

natalieparellano:
These changes look good to me, I am going to test this out with https://github.com/buildpacks/lifecycle/pull/498

image.go
yaelharel: Thank you. I renamed this function and forgot to update the comment.
I'll upload an updated PR soon.

natalieparellano:
Tried testing this out with https://github.com/buildpacks/lifecycle/pull/498, but running into some issues (probably my mistake somewhere). Will look again tomorrow...

natalieparellano:
Verified that `ManifestSize()` outputs the same value that is returned from `docker manifest inspect -v <image>`

natalieparellano (approved):

PR #91, opened 2021-02-26
Adds WithPlatform constructor option for initialization/manifest list querying values

micahyoung:
The changes in this PR are to introduce a `WithPlatform()` option for `NewImage()` to make it easier to override the default `Architecture`/`OS`/`OSVersion` with alternate values in new images and querying manifest lists with `FromBaseImage` or `WithPreviousImage`. Imgutil `NewImage()` currently creates new images with a default of [linux/amd64](https://github.com/buildpacks/imgutil/blob/30601e371ce385a990992de9ed9ff7b373f7c524/remote/remote.go#L108) for `remote` and [daemon OS/amd64](https://github.com/buildpacks/imgutil/blob/30601e371ce385a990992de9ed9ff7b373f7c524/local/local.go#L675) for `local`, and when using `remote.FromBaseImage()` or `remote.WithPreviousImage()`, it queries manifest lists preferring [linux/amd64](https://github.com/google/go-containerregistry/blob/9c81ed02c3b521bea5c56ddc2698f37a788b176f/pkg/v1/remote/options.go#L42) as inherited from `go-containerregistry`. 

The new implementation attempts to be additive and optional, preserving existing behavior (with an exception on Windows¹) but enabling some new scenarios:

* Initialize a new empty image with Architecture/OS/OSVersion:
    `remote`
    ```golang
    img, err := remote.NewImage(
	    "gcr.io/my-new-linux-arm-image:latest",
	    authn.DefaultKeychain,
	    remote.WithPlatform(imgutil.Platform{
		    Architecture: "arm",
		    OS:           "linux",
	    }),
    )
    ...
    img.Architecture() // => "arm"
    img.OS()           // => "linux"
    ```
    `local` with a windows/amd64 daemon
    ```golang
    img, err := local.NewImage(
	    "gcr.io/my-new-windows-arm-image:latest",
	    dockerClient,
	    local.WithPlatform(imgutil.Platform{
		    Architecture: "arm",
		    OS:           "windows",
		    OSVersion:    "10.0.999",
	    }),
    )
    ...
    img.Architecture() // => "arm"
    img.OS()           // => "windows"
    img.OSVersion()    // => "10.0.999"
    ```

* Initialize a new remote image from a base image in multi-arch/os manifest list:
    ```golang
    img, err := remote.NewImage(
	    "gcr.io/my-new-linux-arm-image:latest",
	    authn.DefaultKeychain,
	    remote.FromBaseImage("golang:1.15"), // golang manifest list (contains matching linux/arm image)
	    remote.WithPlatform(imgutil.Platform{
		    Architecture: "arm",
		    OS:           "linux",
	    }),
    )
    ...
    img.Architecture() // => "arm"
    img.OS()           // => "linux"
    ```

* Initialize a new empty image when manifest list doesn't contain image manifest for requested platform:
    ```golang
    img, err := remote.NewImage(
	    "gcr.io/my-new-windows-arm-image:latest",
	    authn.DefaultKeychain,
	    remote.FromBaseImage("golang:1.15"),  // golang manifest list (has no windows/arm image)
	    remote.WithPlatform(imgutil.Platform{
		    Architecture: "arm",
		    OS:           "windows",
	    }),
    )
    ...
    img.Architecture() => "arm"
    img.OS() => "windows"
    ```

There is additional behavior that is intended to stay as close as possible to existing behavior for `NewImage()` and `FromBaseImage()`, but may nevertheless be surprising:

* A single image manifest result for `FromBaseImage()` is always used, even when it doesn't match the `WithPlatform()` values
    ```golang
    img, err := remote.NewImage(
	    "gcr.io/my-supposedly-windows-arm-image:latest",
	    authn.DefaultKeychain,
	    remote.FromBaseImage("amd64/golang:1.15"),  // linux/amd64 image manifest
	    remote.WithPlatform(imgutil.Platform{
		    Architecture: "arm",
		    OS:           "windows",
	    }),
    )
    ...
    img.Architecture() => "amd64"
    img.OS() => "linux"
    ```
    
    ```golang
    // with a windows/amd64 daemon
    img, err := local.NewImage(
	    "gcr.io/my-supposedly-windows-arm-image:latest",
	    authn.DefaultKeychain,
	    remote.FromBaseImage("winamd64/golang:1.15"),  // locally pulled windows/amd64 image
	    remote.WithPlatform(imgutil.Platform{
		    Architecture: "arm",
		    OS:           "windows",
	    }),
    )
    ...
    img.Architecture() => "amd64"
    img.OS() => "windows"
    ```
    
    The reasoning for this behavior is that `WithPlatform()` should primarily be used to assign values to a new image when nothing is inherited - when either no `FromBaseImage()` is given or when the `FromBaseImage()` image is not found. It's secondary purpose is for `remote`'s special edge-case when `FromBaseImage()` needs a manifest list tie-breaker. It shouldn't be treated as a constraint or validator. This approach also follows `ggcr/remote.Image()`, which currently [only uses](https://github.com/google/go-containerregistry/blob/51f01e739161bf35d663240a8ca404777d7d9378/pkg/v1/remote/descriptor.go#L152-L156) it's own `ggcr/remote.WithPlatform()` option when the `ref` is a manifest list, and disregards it for single image manifests even when the platform and image conflict.
    
    In addition to deviating from `ggcr`, the potential alternative approaches I looked into all had downsides:
    * If it returned the base image with the platform values clobbered, it would be difficult for a consumer to figure out what happened and message a user.
    * If it were to ignore the base image result and return an empty image based on the platform values, it would also be difficult to for consumers to deduce what happened since the image ref was otherwise valid.
    * If it were to return an error, we'd need to add custom filtering and error conditions. Not returning an image could potentially prevent consumers from implementing their own filtering/validation after `NewImage` is called. For consumers currently using `NewImage()` and accepting all image platforms, new errors could make backward compatibility challenging, if they only want `WithPlatform` for manifest list tie-breaking and new image defaults.
    
    With the PR approach, consumers can hopefully compare the returned image with their desired platform, and throw an appropriate error if needed. I'm sure there are other considerations though so I'm open for feedback on this.

    
* ¹ New empty Windows images are always initialized with minimal Windows base layers (https://github.com/buildpacks/imgutil/pull/64)

    ```golang
    // with a windows/amd64 daemon
    img, err := local.NewImage(
	    "gcr.io/my-new-windows-arm-image:latest",
	    dockerClient,
	    local.WithPlatform(imgutil.Platform{
		    Architecture: "arm",
		    OS:           "windows",
		    OSVersion:    "10.0.999",
	    }),
    )
    ...
    img.Architecture() // => "arm"
    img.OS()           // => "windows"
    img.OSVersion()    // => "10.0.999"
    len(image.Layers()) // => 1
    ```

    A Windows image cannot be pulled or stored on a daemon without a valid base layer - either from a Microsoft-authored base image or one from `layer.WindowsBaseLayer()`. To simplify new image creation, `NewImage()` now adds one based on `layer.WindowsBaseLayer()`, when `FromBaseImage()` is not used or returns an empty image. This will allow Linux and Windows to behave much more consistently and hopefully hides away this implementation detail for most consumers. It could also allow us to deprecate the shim base layer if it becomes unnecessary.
    
    It is a breaking change for all Windows consumers that already externally add their own base layer with `layer.WindowsBaseLayer()` and `AddLayer()` since that will effectively add it twice. This should be acceptable since it's only known to be used in `pack` which has it behind an experimental flag. Additionally, this new functionality should likely have been included in the original implementation but was blocked on the missing `WithPlatform` support. It is included in the PR since it's the driver for this `WithPlatform` implementation (https://github.com/buildpacks/lifecycle/pull/532), though can be separated out if needed.

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -11,7 +11,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -32,151 +31,120 @@ import (
 type Image struct {
 	docker           client.CommonAPIClient
 	repoName         string
-	platform         imgutil.Platform
 	inspect          types.ImageInspect
 	layerPaths       []string
 	prevImage        *Image // reused layers will be fetched from prevImage
 	downloadBaseOnce *sync.Once
 }
 
-type ImageOption func(*Image) (*Image, error)
-type InitialImageOption ImageOption
+type ImageOption func(*options) error
+
+type options struct {
+	platform          imgutil.Platform
+	baseImageRepoName string
+	prevImageRepoName string
+}
 
 func WithPreviousImage(imageName string) ImageOption {
-	return func(i *Image) (*Image, error) {
-		if _, err := inspectOptionalImage(i.docker, imageName, i.platform); err != nil {
-			return i, err
-		}
-
-		prevImage, err := NewImage(imageName, i.docker, FromBaseImage(imageName))
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to get previous image '%s'", imageName)
-		}
-		i.prevImage = prevImage
-
-		return i, nil
+	return func(i *options) error {
+		i.prevImageRepoName = imageName
+		return nil
 	}
 }
 
 func FromBaseImage(imageName string) ImageOption {
-	return func(i *Image) (*Image, error) {
-		var (
-			err     error
-			inspect types.ImageInspect
-		)
-
-		if inspect, err = inspectOptionalImage(i.docker, imageName, i.platform); err != nil {
-			return i, err
-		}
-
-		i.inspect = inspect
-		i.layerPaths = make([]string, len(i.inspect.RootFS.Layers))
-
-		return i, nil
+	return func(i *options) error {
+		i.baseImageRepoName = imageName
+		return nil
 	}
 }
 
-func WithPlatform(platform imgutil.Platform) InitialImageOption {
-	return func(i *Image) (*Image, error) {
-		if platform.OS != "" && platform.OS != i.inspect.Os {
-			return nil, fmt.Errorf(`invalid os: platform os "%s" must match the daemon os "%s"`, platform.OS, i.inspect.Os)
-		}
-
-		i.inspect.Architecture = platform.Architecture
-		i.inspect.OsVersion = platform.OSVersion
-
+func WithPlatform(platform imgutil.Platform) ImageOption {
+	return func(i *options) error {
 		i.platform = platform
-
-		return i, nil
+		return nil
 	}
 }
 
-func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...interface{}) (*Image, error) {
-	var err error
+func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...ImageOption) (*Image, error) {
+	imageOpts := &options{}
+	for _, op := range ops {
+		if err := op(imageOpts); err != nil {
+			return nil, err
+		}
+	}
 
-	defaultPlatform, err := defaultPlatform(dockerClient)
+	platform, err := defaultPlatform(dockerClient)
 	if err != nil {
 		return nil, err
 	}
+	daemonOS := platform.OS
 
-	inspect := defaultInspect(defaultPlatform)
+	if (imageOpts.platform != imgutil.Platform{}) {
+		platform = imageOpts.platform
+	}
+
+	if imageOpts.platform.OS != "" && imageOpts.platform.OS != daemonOS {
+		return nil, fmt.Errorf(`invalid os: platform os "%s" must match the daemon os "%s"`, imageOpts.platform.OS, platform.OS)
+	}
+
+	inspect := defaultInspect(platform)
 
 	image := &Image{
 		docker:           dockerClient,
 		repoName:         repoName,
 		inspect:          inspect,
-		platform:         defaultPlatform,
 		layerPaths:       make([]string, len(inspect.RootFS.Layers)),
 		downloadBaseOnce: &sync.Once{},
 	}
 
-	image, err = processImageOptions(image, ops)
-	if err != nil {
-		return nil, err
+	if imageOpts.prevImageRepoName != "" {
+		// with previous image
+		if _, err := inspectOptionalImage(dockerClient, imageOpts.prevImageRepoName, platform); err != nil {
+			return nil, err
+		}
+
+		prevImage, err := NewImage(imageOpts.prevImageRepoName, dockerClient, FromBaseImage(imageOpts.prevImageRepoName))
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get previous image '%s'", repoName)
+		}
+		image.prevImage = prevImage
 	}
 
-	image, err = prepareImage(image)
-	if err != nil {
-		return nil, err
+	if imageOpts.baseImageRepoName != "" {
+		if inspect, err = inspectOptionalImage(dockerClient, imageOpts.baseImageRepoName, platform); err != nil {
+			return nil, err
+		}
+
+		image.inspect = inspect
+		image.layerPaths = make([]string, len(image.inspect.RootFS.Layers))
 	}
 
-	return image, nil
-}
-
-func processImageOptions(image *Image, ops []interface{}) (*Image, error) {
-	sort.Slice(ops, func(i, _ int) bool {
-		switch ops[i].(type) {
-		case InitialImageOption:
-			return true
-		default:
-			return false
-		}
-	})
-
-	for _, op := range ops {
-		var err error
-
-		switch opFn := op.(type) {
-		case InitialImageOption:
-			image, err = opFn(image)
-		case ImageOption:
-			image, err = opFn(image)
-		}
+	if image.inspect.Os == "windows" && len(image.inspect.RootFS.Layers) == 0 {
+		layerReader, err := layer.WindowsBaseLayer()
 		if err != nil {
 			return nil, err
 		}
+
+		layerFile, err := ioutil.TempFile("", "imgutil.local.image.windowsbaselayer")
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to create temp file")
+		}
+		defer layerFile.Close()
+
+		hasher := sha256.New()
+
+		multiWriter := io.MultiWriter(layerFile, hasher)
+
+		if _, err := io.Copy(multiWriter, layerReader); err != nil {
+			return nil, errors.Wrap(err, "failed to copy base layer")
+		}
+
+		diffID := "sha256:" + hex.EncodeToString(hasher.Sum(nil))
+
+		image.inspect.RootFS.Layers = append(image.inspect.RootFS.Layers, diffID)
+		image.layerPaths = append(image.layerPaths, layerFile.Name())
 	}
-	return image, nil
-}
-
-func prepareImage(image *Image) (*Image, error) {
-	if image.inspect.Os != "windows" || len(image.inspect.RootFS.Layers) != 0 {
-		return image, nil
-	}
-
-	layerReader, err := layer.WindowsBaseLayer()
-	if err != nil {
-		return nil, err
-	}
-
-	layerFile, err := ioutil.TempFile("", "imgutil.local.image.windowsbaselayer")
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to create temp file")
-	}
-	defer layerFile.Close()
-
-	hasher := sha256.New()
-
-	multiWriter := io.MultiWriter(layerFile, hasher)
-
-	if _, err := io.Copy(multiWriter, layerReader); err != nil {
-		return nil, errors.Wrap(err, "failed to copy base layer")
-	}
-
-	diffID := "sha256:" + hex.EncodeToString(hasher.Sum(nil))
-
-	image.inspect.RootFS.Layers = append(image.inspect.RootFS.Layers, diffID)
-	image.layerPaths = append(image.layerPaths, layerFile.Name())
 
 	return image, nil
 }
--- a/local/local.go
+++ b/local/local.go
@@ -87,9 +87,11 @@ func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...Image
 		return nil, err
 	}
 
-	platform, err = processPlatformOption(platform, imageOpts.platform)
-	if err != nil {
-		return nil, err
+	if (imageOpts.platform != imgutil.Platform{}) {
+		if err := validatePlatformOption(platform, imageOpts.platform); err != nil {
+			return nil, err
+		}
+		platform = imageOpts.platform
 	}
 
 	inspect := defaultInspect(platform)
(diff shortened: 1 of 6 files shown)


local/local_test.go
```diff
@@ -1266,8 +1403,8 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 			inspect, _, err := dockerClient.ImageInspectWithRaw(context.TODO(), prevName)
 			h.AssertNil(t, err)
 
-			prevLayer1SHA = inspect.RootFS.Layers[0]
-			prevLayer2SHA = inspect.RootFS.Layers[1]
+			prevLayer1SHA = inspect.RootFS.Layers[len(inspect.RootFS.Layers)-2]
+			prevLayer2SHA = inspect.RootFS.Layers[len(inspect.RootFS.Layers)-1]
```
natalieparellano, line 1407: Is this to avoid having the assertion switch on the daemon OS?

local/local.go
```diff
@@ -71,28 +76,107 @@ func FromBaseImage(imageName string) ImageOption {
...
+	image, err = prepareImage(image)
+	if err != nil {
+		return nil, err
+	}
+
+	return image, nil
+}
+
+func processImageOptions(image *Image, ops []interface{}) (*Image, error) {
+	sort.Slice(ops, func(i, _ int) bool {
```
natalieparellano, line 127: Is this to ensure that `WithPlatform` is always applied first?

natalieparellano:
@micahyoung I think the flow you proposed makes sense. I know @ekcasey will want to review this one, but this looks good to me!

local/local_test.go
```diff
@@ -84,91 +84,236 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
-					expectedOSVersion := ""
-					if daemonOS == "windows" {
-						// this nanoserver windows/arm image exists and pulls. Not sure whether it works for anything else.
-						armBaseImageName = "mcr.microsoft.com/windows/nanoserver@sha256:29e2270953589a12de7a77a7e77d39e3b3e9cdfd243c922b3b8a63e2d8a71026"
-						expectedArmArch = "arm"
-						expectedOSVersion = "10.0.17763.1040"
-					}
+			when("#WithPlatform", func() {
+				when("base image and platform architecture/OS do not match", func() {
+					it("uses the base image architecture/OS, ignoring platform", func() {
```
ekcasey, line 245: Should we just fail in this case? Given that we can't sensibly reconcile the two options?

local/local.go
```diff
@@ -71,28 +76,107 @@ func FromBaseImage(imageName string) ImageOption {
...
+		i.inspect.Architecture = platform.Architecture
+		i.inspect.OsVersion = platform.OSVersion
+
+		i.platform = platform
+
+		return i, nil
+	}
+}
+
+func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...interface{}) (*Image, error) {
```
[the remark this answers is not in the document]
ekcasey: This is getting complicated enough that we should definately add a doc string describing the behavior of this function, regardless of whether we make the change suggested above.

local/local_test.go
```diff
@@ -84,91 +84,236 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
-					armBaseImageName := "arm64v8/busybox@sha256:50edf1d080946c6a76989d1c3b0e753b62f7d9b5f5e66e88bef23ebbd1e9709c"
-					expectedArmArch := "arm64"
-					expectedOSVersion := ""
-					if daemonOS == "windows" {
-						// this nanoserver windows/arm image exists and pulls. Not sure whether it works for anything else.
-						armBaseImageName = "mcr.microsoft.com/windows/nanoserver@sha256:29e2270953589a12de7a77a7e77d39e3b3e9cdfd243c922b3b8a63e2d8a71026"
-						expectedArmArch = "arm"
-						expectedOSVersion = "10.0.17763.1040"
-					}
+			when("#WithPlatform", func() {
```
ekcasey, line 243: Do we need a set of tests like this the test the combination of `WithPlatform` and `WithPreviousImage`?

remote/remote.go
```diff
@@ -71,50 +98,119 @@ func NewImage(repoName string, keychain authn.Keychain, ops ...ImageOption) (*Im
...
 
-func newV1Image(keychain authn.Keychain, repoName string) (v1.Image, error) {
+func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platform) (v1.Image, error) {
 	ref, auth, err := referenceForRepoName(keychain, repoName)
 	if err != nil {
 		return nil, err
 	}
 
-	image, err := remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport))
+	image, err := remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport), remote.WithPlatform(v1.Platform(platform)))
```
ekcasey, line 176: Did we ensure that this still works if you are fetching a windows image (not from a manifest list) without `WithPlatform`? I am assuming it does after looking at the GGCR code, but just double checking.

local/local.go
```diff
@@ -664,16 +748,24 @@ func inspectOptionalImage(docker client.CommonAPIClient, imageName string) (type
...
 }
 
-func defaultInspect(docker client.CommonAPIClient) (types.ImageInspect, error) {
-	daemonInfo, err := docker.Info(context.Background())
+func defaultInspect(platform imgutil.Platform) types.ImageInspect {
+	return types.ImageInspect{
+		Os:           platform.OS,
+		Architecture: platform.Architecture,
+		OsVersion:    platform.OSVersion,
+		Config:       &container.Config{},
```
ekcasey, line 756: I we including fields like `Variant` and `Features` in `imgutil.Platform` and then ignoring them when they are set. Maybe we should just remove them?

remote/remote.go
```diff
@@ -24,17 +27,19 @@ import (
 type Image struct {
 	keychain   authn.Keychain
 	repoName   string
+	platform   imgutil.Platform
```
ekcasey, line 30: It feels a little weird to me that we have this field in the Image struct that could be inaccurate. That would happen if
1. A user called `SetOS`
2. No platform was provided but the base image is a windows image

Someone could easily introduce a bug if they assume this is always set correctly.

local/local.go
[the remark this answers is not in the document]
ekcasey: Potentially better idea to replace the interfaces!

Let's copy GGCR and make our `ImageOption` mutate an internal options struct, instead of directly mutating the image. We can set options fields like `baseImage` and `platform` and then do sane things with the results once all the options are applied.

https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/options.go#L28-L40

This also solves the problem with the platform field described in a different comment b/c we wouldn't need to save the platform as separate field on the Image.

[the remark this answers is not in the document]
micahyoung: Sounds good! I'll change the `ImageOption`s to not modify the image, so I can make them all the same type again.

And I'll add the missing doc strings for them, including the caveats from this PR thread - totally missed that one

local/local.go
micahyoung: Yes, though as per Emily's [comment](https://github.com/buildpacks/imgutil/pull/91/files/038872784f71cf5d70fce08b92ebf5716337180d#r583956120), I think I'll add another commit to this to move the processing of all the options into here.

local/local_test.go
micahyoung: I could be swayed but it felt like it starts to deviate from `remote`, which in this impl just wraps `ggcr`'s [behavior](https://github.com/google/go-containerregistry/blob/51f01e739161bf35d663240a8ca404777d7d9378/pkg/v1/remote/descriptor.go#L152-L156), which is to only consider platform for manifest list refs (which are never a thing for `local` since it's refs are unique).

My general take on `WithPlatform()` had been primarily to assign values to a new image when nothing is inherited - when either no `FromBaseImage()` is given or when the `FromBaseImage()` image is not found. It's secondary purpose is for remote's special edge-case when FromBaseImage() needs a manifest list tie-breaker. I wasn't yet treating it as a constraint or validator.

local/local_test.go
micahyoung: Yes, and I realize it's repeated in a couple other places - I'll toss in a helper to clarify it. 

The reason it needed to be changed here is that `NewImage()` can no longer create Windows images with no layers at all (which technically *could* be saved to a daemon, but have no value that I could see).

If layerless Windows images are important, I can certainly take a shot at getting them back

remote/remote.go
micahyoung: Since `descriptor.Image()` (called by `remote.Image()` [currently ignores](https://github.com/google/go-containerregistry/blob/51f01e739161bf35d663240a8ca404777d7d9378/pkg/v1/remote/descriptor.go#L152-L156) the platform for any non-manifest-list result. 

I think the [first remote/#NewImage/#BaseImage](https://github.com/buildpacks/imgutil/blob/038872784f71cf5d70fce08b92ebf5716337180d/remote/remote_test.go#L143) test kinda covers this, since it doesn't match ggcrs `linux/amd64` default platform. I can change that test, or add another that deviates from all the default platform fields.

micahyoung:
Thanks for the reviews! I'm gonna add another commit to address these:
[ ] Add doc comments each `ImageOption`
[ ] Remove `Variants` and `Features` from `imgutil.Platform`
[ ] Refactor `ImageOption` to not alter the image directly, restore `ImageOption` instead of `interface{}` in constructor
[ ] Sanity check workflow of when `WithPlatform` + `SetOS` are used.
[ ] Add test helper for assertions on top-2-image layers

local/local_test.go
micahyoung: I think this gets tricky because we have `WithPlatform` doing double-duty for `remote` (1. as a new image value setter, 2. as a tie-breaker for manifest lists). If we don't want `imgutil` consumers to need to know ahead of time wether a remote ref will be a manifest list or image manifest, and swap in our out `WithPlatform` accordingly, I think we need to keep passing-through image manifests so consumers can decide what to do to with them. And because we're stuck with that for `remote`, I think we need to stay consistent in `local`.

I could be swayed though, happy to hear counter arguments.

local/local.go
```diff
@@ -71,28 +76,107 @@ func FromBaseImage(imageName string) ImageOption {
 	}
 }
 
-func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...ImageOption) (*Image, error) {
+func WithPlatform(platform imgutil.Platform) InitialImageOption {
```
jabrown85, line 79: What do you think about `WithDefaultPlatform` or `WithPlatformDefault`? I was thinking that the `Default` might convey this wouldn't be used in the scenario where you are using a previous image.

But maybe after Emily's comments/suggestions - this is no longer needed.

micahyoung: I like this renaming suggestion, assuming we keep the behavior as it is. `WithDefaultPlatform` feels best to me, as it follows proposition-adjective-noun pattern of `FromBaseImage`/`FromPreviousImage`.

remote/remote.go
micahyoung: I removed `platform` from the `Image`, as suggested, with the latest set of commits.

remote/remote.go
micahyoung: Added new case: https://github.com/buildpacks/imgutil/pull/91/files#diff-55e016054597444d1031ab5c620b77bdf8118f0ee1f82888a2b851850e151726R171

local/local_test.go
micahyoung: New test cases: 
local: https://github.com/buildpacks/imgutil/pull/91/files#diff-088ed6061849ec1bfbe08e88f1b8da21e3569421ce717bb8b01e4b94e5523c63R322
remote: https://github.com/buildpacks/imgutil/pull/91/files#diff-55e016054597444d1031ab5c620b77bdf8118f0ee1f82888a2b851850e151726R439

local/local.go
micahyoung: Removed all fields besides `Architecture`/`OS`/`OSVersion`

local/local_test.go
micahyoung: I [added](https://github.com/buildpacks/imgutil/pull/91/files#diff-fd2fec5210d5a5d51c6442448b9a8577ad13498d133c4d9c4839ce292667edb5R359) a new helper that works as follows:

from:
```golang
prevLayer1SHA = inspect.RootFS.Layers[len(inspect.RootFS.Layers)-2]
```
to:
```golang
// negative index (offset from len(inspect.RootFS.Layers))
prevLayer1SHA = h.StringElementAt(inspect.RootFS.Layers), -2)

// also normal index
prevLayer1SHA = h.StringElementAt(inspect.RootFS.Layers), 1)
```

micahyoung:
After the above changes, plus some manual and [newly enabled](https://github.com/buildpacks/lifecycle/pull/541) acceptance lifecycle testing on Windows, I'm feeling like this is in a good state and ready for another look, if you all wouldn't mind.

local/local.go
```diff
@@ -87,13 +102,98 @@ func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...Image
...
+	if err != nil {
+		return image, err
+	}
+
+	image, err = processBaseImageOption(image, imageOpts.baseImageRepoName, platform, dockerClient)
+	if err != nil {
+		return image, err
+	}
+
+	image, err = prepareNewWindowsImage(image)
```
ekcasey, line 115: It feels a bit weird to read through this function and see that `prepareNewWindowsImage` is always called. Maybe we want to extract the guard conditionals out of the `prepare*` functions. E.G
```
if image.inspect.Os == "windows" {
    image, err = prepareNewWindowsImage(image)
    if err != nil {
        return nil, err
    }
}
```

local/local.go
```diff
@@ -87,13 +102,98 @@ func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...Image
...
 	}
 
-	for _, v := range ops {
-		image, err = v(image)
-		if err != nil {
-			return nil, err
-		}
+	image, err = processPreviousImageOption(image, imageOpts.prevImageRepoName, platform, dockerClient)
+	if err != nil {
+		return image, err
```
ekcasey, line 107:
```suggestion
		return nil, err
```
We typically return empty values for all other return values, when error in non-nil

local/local.go
```diff
@@ -87,13 +102,98 @@ func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...Image
...
+	}
+
+	image, err = processBaseImageOption(image, imageOpts.baseImageRepoName, platform, dockerClient)
+	if err != nil {
+		return image, err
```
ekcasey, line 112:
```suggestion
		return nil, err
```

local/local.go
```diff
@@ -87,13 +102,98 @@ func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...Image
...
+	}
+
+	image, err = prepareNewWindowsImage(image)
+	if err != nil {
+		return image, err
```
ekcasey, line 117:
```suggestion
		return nil, err
```

local/local.go
```diff
@@ -87,13 +102,98 @@ func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...Image
...
+	}
+
+	prevImage, err := NewImage(prevImageRepoName, dockerClient, FromBaseImage(prevImageRepoName))
+	if err != nil {
+		return image, errors.Wrapf(err, "failed to get previous image '%s'", prevImageRepoName)
+	}
+
+	image.prevImage = prevImage
+
+	return image, nil
```
ekcasey, line 151: Because `image` is a pointer we can set `image.prevImage` without returning the modified `image`. I think we should remove the first result parameter to make this clear. Without reading the implementation the function signature would seem to imply that the provided `image` has not been mutated.

local/local.go
```diff
@@ -87,13 +102,98 @@ func NewImage(repoName string, dockerClient client.CommonAPIClient, ops ...Image
...
+	multiWriter := io.MultiWriter(layerFile, hasher)
+
+	if _, err := io.Copy(multiWriter, layerReader); err != nil {
+		return image, errors.Wrap(err, "failed to copy base layer")
+	}
+
+	diffID := "sha256:" + hex.EncodeToString(hasher.Sum(nil))
+
+	image.inspect.RootFS.Layers = append(image.inspect.RootFS.Layers, diffID)
+	image.layerPaths = append(image.layerPaths, layerFile.Name())
```
ekcasey, line 196:
```suggestion
	image.AddLayerWithDiffID(layerFile.Name(), diffID)
```
I think this is slightly more robust to implementation changes.

ekcasey (commented):
Looking good! Just a couple suggestions for cleanup.

local/local.go
micahyoung: Yeah, I was feeling the same thing. The prior private `prepare*` functions have internal conditional guards too. Is it ok to have Windows be the weird one or should I look for a pattern to move the guards out of all of them?

micahyoung: Yeah, good call. I'll take a pass at it. The `prepare*` function guards should be easy to move out.

local/local.go
micahyoung: Now fixed in local and remote

local/local.go
micahyoung: I moved all the guards up into `NewImage` for local and remote.

local/local.go
micahyoung: I changed the function signatures in local and remote

local/local.go
micahyoung: I only ended up doing this in local, as remote was able to skip the temp file and write directly from WindowsBaseLayer to the registry.

local/local.go
micahyoung: WithDefaultPlatform feels like a good fit for the current behavior.

micahyoung:
I made changes based on the feedback, which did clarify the logic. I feel like I'll want to revisit this as some point after living with it a bit and maybe looking into some different option pattern examples - I feel there must be some tidier ways to move the logic back into the image options, but still enforce the order of execution that we need.

local/local_test.go
ekcasey: > The reason it needed to be changed here is that NewImage() can no longer create Windows images with no layers at all (which technically could be saved to a daemon, but have no value that I could see).

This statement confuses me. I thought we were always adding a base layer on windows images because it is impossible to pull a windows image without one. Can you successfully docker load an empty image, even though you can't pull one?

ekcasey (approved):

natalieparellano (approved):

PR #95, opened 2021-03-07
Changes registry test helper to support non-localhost

micahyoung:
This is useful for consumers of DockerRegistry helper, particularly lifecycle (maybe pack, whose [impl](https://github.com/buildpacks/pack/blob/main/testhelpers/registry.go#L182) this was based on and already supports this), when running test containers that access the separate registry container, listening on a published host port. Currently, this was only possible with `--network=host` which can't be used on Windows. This also lets a Unix workstation run tests against a remote Windows (or any arch/OS) daemon with `DOCKER_HOST=tcp://<daemon host>` set.

The test registry now has a `Host` field, whose value is autodetected from the host and which can be used in image refs that need to be written to it: ex: `registry.Host + ":" + registry.Port + "/image-name:tag"`.

The logic to detect the host IP:
  - Any exact IPs (`<host IP>/32`) in `insecure-registry` entries, assume that's the host's IP
  - ... or, fallback to "localhost"

This does not change the current behavior for imgutil when running on GHA, which has none of the above set so will use "localhost".

This also replaces `stefanscherer/registry-windows`, which had a bug where it would return sporadic 503s. This bug appears fixed in docker-distribution 2.7.1 but the image hasn't been update for over 2 years. This is replaced with a newly created [`micahyoung/registry`](https://github.com/micahyoung/registry-image) which hosts equivalent Windows & Linux AMD64/ARM images in a single manifest list and is updated automatically through GHA. This should make the test registry behave as consistently as possible, regardless of the OS/Arch. I'd be happy to donate this to `buildpacks` org or even move the `Dockerfile`s into this repo to be built on-demand.

(updated, remove extra registry host detection logic for simplicity)

### DIFF
--- a/remote/remote_test.go
+++ b/remote/remote_test.go
@@ -8,6 +8,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-containerregistry/pkg/name"
+
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
@@ -17,7 +19,7 @@ import (
 	h "github.com/buildpacks/imgutil/testhelpers"
 )
 
-var dockerRegistry *h.DockerRegistry
+var dockerRegistry, readonlyDockerRegistry *h.DockerRegistry
 
 func newTestImageName(providedPrefix ...string) string {
 	prefix := "pack-image-test"
@@ -35,9 +37,15 @@ func TestRemote(t *testing.T) {
 	h.AssertNil(t, err)
 	defer os.RemoveAll(dockerConfigDir)
 
-	dockerRegistry = h.NewDockerRegistryWithAuth(dockerConfigDir)
+	sharedStorageOpt := h.WithSharedStorageVolume("test-registry-volume" + h.RandString(10))
+
+	dockerRegistry = h.NewDockerRegistry(h.WithAuth(dockerConfigDir), sharedStorageOpt)
 	dockerRegistry.Start(t)
 	defer dockerRegistry.Stop(t)
+
+	readonlyDockerRegistry = h.NewDockerRegistry(sharedStorageOpt)
+	readonlyDockerRegistry.Start(t)
+	defer readonlyDockerRegistry.Stop(t)
 
 	os.Setenv("DOCKER_CONFIG", dockerRegistry.DockerDirectory)
 	defer os.Unsetenv("DOCKER_CONFIG")
@@ -76,6 +84,13 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				arch, err := img.Architecture()
 				h.AssertNil(t, err)
 				h.AssertEq(t, arch, "amd64")
+			})
+
+			it("fails to save to read-only registry", func() {
+				readOnlyRepoName := readonlyDockerRegistry.RepoName("pack-image-test" + h.RandString(10))
+				img, err := remote.NewImage(readOnlyRepoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+				h.AssertError(t, img.Save(), "Method Not Allowed")
 			})
 		})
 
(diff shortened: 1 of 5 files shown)


micahyoung:
For reference, lifecycle Draft PR (with analyzer tests enabled for Windows): https://github.com/buildpacks/lifecycle/pull/541

ekcasey:
Looks good, but I do worry a touch about making it so easy to open up a totally unauthenticated registry on a public network. The analyzer tests specifically tests scenarios when auth is not required to pull images. Could we do either or both
* Make the registry created with `NewDockerRegistry` require auth to write but not to read?
* Rename `NewDockerRegistry` to something like `NewDockerRegistryWithoutAuth` (or `NewDockerRegistryWithPublicRead` if we do the bullet point above) and add a doc comment telling folks to be careful if/when they use this that they have other network-level protections in place.

micahyoung:
@ekcasey Great points. I'll take a pass a public-read/authenticated-write as the overridable defaults (with very explicit warnings).

micahyoung:
I updated this with the readwrite-with-auth/readonly-without-auth approach and added some (hopefully) useful test coverage to ensure these get helpers get exercised. I added registry helper options for auth and shared-volumes.  I bumped https://github.com/buildpacks/lifecycle/pull/541 to show how the dual-registry-shared-data works and everything is green.

testhelpers/testhelpers.go
```diff
@@ -177,8 +179,25 @@ func DockerRmi(dockerCli dockercli.CommonAPIClient, repoNames ...string) error {
 	return err
 }
 
-func PushImage(dockerCli dockercli.CommonAPIClient, ref string) error {
-	rc, err := dockerCli.ImagePush(context.Background(), ref, dockertypes.ImagePushOptions{RegistryAuth: "{}"})
+func PushImage(dockerCli dockercli.CommonAPIClient, refStr string) error {
+	ref, err := name.ParseReference(refStr, name.WeakValidation)
+	if err != nil {
+		return err
+	}
```
natalieparellano, line 186:
```suggestion
	}
	
```

testhelpers/docker_registry.go
```diff
@@ -103,34 +159,85 @@ func (r *DockerRegistry) Start(t *testing.T) {
...
 func (r *DockerRegistry) RepoName(name string) string {
-	return "localhost:" + r.Port + "/" + name
+	return r.Host + ":" + r.Port + "/" + name
 }
 
 func (r *DockerRegistry) EncodedLabeledAuth() string {
 	return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(`{"username":"%s","password":"%s"}`, r.username, r.password)))
 }
 
+func DockerHostname(t *testing.T) string {
```
natalieparellano, line 202: Would it be worth adding a docs comment here describing the chain of precedence?

testhelpers/docker_registry.go
```diff
@@ -73,17 +95,51 @@ func (r *DockerRegistry) Start(t *testing.T) {
...
+	} else {
+		// make read-only without auth
+		readOnlyEnv := `REGISTRY_STORAGE_MAINTENANCE_READONLY={"enabled":true}`
+		registryEnv = append(registryEnv, readOnlyEnv)
+	}
+
+	var volumeBinds []string
+	var containerUser string
+	if r.volumeName != "" {
+		// try to create volumes that do not exist
```
natalieparellano, line 107:
```suggestion
		// try to create volumes that may exist
```

remote/remote_test.go
```diff
@@ -1027,6 +1042,33 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			it("can be pulled from unauthenticated registry", func() {
+				img, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+
+				err = img.SetLabel("mykey", "newValue")
+				h.AssertNil(t, err)
+
+				h.AssertNil(t, img.Save())
+
+				name, err := name.ParseReference(repoName, name.WeakValidation)
```
natalieparellano, line 1054: I'm a bit confused... how does this work? It looks like `repoName` is on the authenticated registry. Where are we pulling from the unauthenticated one?

micahyoung: Yeah, this could use some clarification. What it attempts to do is:
1. Write a new image on the auth'd registry with ref like: `localhost:5555/my-image-1234`
2. Parse out the name out of the auth'd ref: `my-image.1234`
3. Rebuild the ref string for the same image on the unauth'd repo: `localhost:6666/my-image-1234`
4. Show that we can pull that without any credentials (since we never added any for `localhost:6666` to the `DOCKER_CONFIG`).

micahyoung: I'll take another pass and see if I can clean it up. TBH, the test is mostly here for some coverage for our `WithSharedStorageVolume` helper, theoretically unauthenticated registries are covered with the DockerHub examples.

testhelpers/testhelpers.go
```diff
@@ -177,8 +179,26 @@ func DockerRmi(dockerCli dockercli.CommonAPIClient, repoNames ...string) error {
 	return err
 }
 
-func PushImage(dockerCli dockercli.CommonAPIClient, ref string) error {
-	rc, err := dockerCli.ImagePush(context.Background(), ref, dockertypes.ImagePushOptions{RegistryAuth: "{}"})
+//PushImage pushes and image to a registry, optionally using credentials from any set DOCKER_CONFIG
```
ekcasey, line 182:
```suggestion
//PushImage pushes an image to a registry, optionally using credentials from any set DOCKER_CONFIG
```

ekcasey (approved):
This looks fantastic!

natalieparellano (approved):

PR #96, opened 2021-03-12
Retry fetching images on EOF errors

natalieparellano:
Based on @dwillist 's comment here: https://github.com/buildpacks/lifecycle/issues/414#issuecomment-796779364 I think this is one way forward. I am not totally sure of the approach, would love to get feedback.

Signed-off-by: Natalie Arellano <narellano@vmware.com>


remote/remote.go
```diff
@@ -186,18 +186,24 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
...
-	if err != nil {
-		if transportErr, ok := err.(*transport.Error); ok && len(transportErr.Errors) > 0 {
-			switch transportErr.StatusCode {
-			case http.StatusNotFound, http.StatusUnauthorized:
+	var image v1.Image
+	for i := 0; i < 3; i++ {
+		image, err = remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport), remote.WithPlatform(v1Platform))
+		if err != nil {
+			if err == io.EOF {
+				continue // retry if EOF
```
aemengo, line 194: Given the nature of the linked comment, I'm wondering if we should introduce a little delay between attempts.

natalieparellano: Added in https://github.com/buildpacks/imgutil/pull/96/commits/36099cee419a40551bac14161c75e57b3b0ce293

jabrown85 (approved):

aemengo (approved):

PR #97, opened 2021-03-20
Fix docker windows 20.04 tests

micahyoung:
This fixes the recent change in GHA default runners from Docker 19.03 to 20.10 ([docs change](https://github.com/actions/virtual-environments/commit/faee73e79a5bdbf2f9d9d79c81453dfcf3452485#diff-816219d3c9f30fcc1c40f88ca826c1662bf99eded781c1b419dd2dfb577ae227R58), the change to the worker happened last week)

This version of the daemon now returns additional text in non-error responses that can't be parsed by our version of the docker client library. Instead the test helper now ignores any parsing failures on only returns on API function errors.


testhelpers/testhelpers.go
```diff
@@ -397,13 +397,9 @@ func checkResponseError(r io.Reader) error {
 	for {
 		decoder := json.NewDecoder(r)
 		var jsonMessage jsonmessage.JSONMessage
-		err := decoder.Decode(&jsonMessage)
-		if err == io.EOF {
+		if err := decoder.Decode(&jsonMessage); err != nil {
```
jabrown85, line 400: What does a "bad" message look like?

micahyoung: I'm still not 100% sure. I couldn't repro locally and the closest I got was this [error](https://github.com/buildpacks/imgutil/runs/2114142490?check_suite_focus=true#step:4:136) in the test run, on what otherwise would have been green:

```
image_test.go:32: Expected nil: parsing daemon response: json: cannot unmarshal string into Go value of type jsonmessage.JSONMessage
```

The exact same commit, on the PR test run run before it was merged to main (but which just happened to run on the older daemon), was successful [here](https://github.com/buildpacks/imgutil/runs/2112977347?check_suite_focus=true#step:4:135)

natalieparellano: Hmm... could we read all the bytes from `r` just to see what it is?

micahyoung: @jabrown85 @natalieparellano I rewrote `checkResponseError` to try to output the full error on a parsing failure. I couldn't repro the original failure scenario to see what the original message was but this should catch it (or perhaps fixed the error)

natalieparellano (approved):

testhelpers/testhelpers.go
natalieparellano: I want to say that this was fixed on the runner's end - as https://github.com/buildpacks/imgutil/pull/96 is now passing on Windows after seeing this failure yesterday. But these changes are still an improvement on what we had before!

jabrown85 (approved):

testhelpers/testhelpers.go
natalieparellano: Uggh... this appears to be a non-deterministic problem: https://github.com/buildpacks/imgutil/pull/98/checks?check_run_id=2177275953

micahyoung: Hmm, I wonder if this PR's implementation has the same issue. I'm going to re-run the tests a couple times

natalieparellano:
Since this is green, I'm going to try updating & merging it. Fingers crossed...

PR #99, opened 2021-03-23
Fix return value for 3 EOF errors

natalieparellano:
If we get 3 EOF errors, we want to return a nil image and an error.
In the former implementation, we would fall through to the return statement at the end of the function (nil error).

Thanks to @yaelharel for pointing this out.

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -23,6 +23,8 @@ import (
 	"github.com/buildpacks/imgutil"
 )
 
+const maxRetries = 2
+
 type Image struct {
 	keychain   authn.Keychain
 	repoName   string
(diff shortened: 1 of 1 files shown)


remote/remote.go
```diff
@@ -187,11 +189,11 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
 	}
 
 	var image v1.Image
-	for i := 0; i < 3; i++ {
+	for i := 0; i < maxRetries+1; i++ {
```
yaelharel, line 192:
```suggestion
	for i := 0; i <= maxRetries; i++ {
```

yaelharel (approved):
Great!
Added a small suggestion. Feel free to ignore it :)

jabrown85 (approved):

remote/remote.go
natalieparellano: I like it!

PR #102, opened 2021-04-23
Fixes registry helper on GHA

micahyoung:
It seems with Docker 20.10.6 ([release notes](https://docs.docker.com/engine/release-notes/#20106)), multiple host port mappings are now returned for a single container port (one host mapping for IPv4, a second for IPv6):

Printf of `ContainerInspect().NetworkSettings.Ports["5000/tcp"]`:
```
[]nat.PortBinding{
  nat.PortBinding{
    HostIP:"0.0.0.0", HostPort:"49160"
  },
  nat.PortBinding{
    HostIP:"::", HostPort:"49160"}
  }
}
```
The previous implementation only looked for a single entry during the wait logic. I change it here to now loop through all mappings, choosing the IPv4 mapping (when IP is `0.0.0.0`), and returning it's port. If nothing is found, fail the test with a better error, to avoid apparent 10s timeout message from before.

### DIFF
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -155,12 +155,17 @@ func (r *DockerRegistry) Start(t *testing.T) {
 		AssertNil(t, err)
 
 		hostPortMap := inspect.NetworkSettings.Ports["5000/tcp"]
-		if len(hostPortMap) == 1 {
-			r.Port = hostPortMap[0].HostPort
-			break
+		for _, hostPortEntry := range hostPortMap {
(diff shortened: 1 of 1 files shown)


micahyoung:
This also should fix `lifecycle`, which is using this helper and failing on main for the same reason: https://github.com/buildpacks/lifecycle/runs/2421388413?check_suite_focus=true

jabrown85 (approved):

natalieparellano (approved):

PR #106, opened 2021-05-04
Change file permissions

soniasingla:

### DIFF
--- a/tools/bcdhive_generator/bcdhive_hivex.go
+++ b/tools/bcdhive_generator/bcdhive_hivex.go
@@ -80,7 +80,7 @@ func HiveBCD() ([]byte, error) {
 		return nil, err
 	}
 
-	if err := ioutil.WriteFile(hiveFile.Name(), origHiveBytes, 0666); err != nil {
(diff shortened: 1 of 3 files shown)


tools/bcdhive_generator/bcdhive_hivex.go
```diff
@@ -80,7 +80,7 @@ func HiveBCD() ([]byte, error) {
 		return nil, err
 	}
 
-	if err := ioutil.WriteFile(hiveFile.Name(), origHiveBytes, 0666); err != nil {
+	if err := ioutil.WriteFile(hiveFile.Name(), origHiveBytes, 0600); err != nil {
```
natalieparellano, line 83: @micahyoung do you foresee any issues here? Lifecycle acceptance tests are passing with this branch of imgutil.

micahyoung: Looks good to me... the `go generate` [ran](https://github.com/buildpacks/imgutil/runs/2504046859?check_suite_focus=true#step:4:166) which is what ends up reading that file, so it appears to be fine

natalieparellano (approved):

PR #109, opened 2021-05-04
Add comment to ignore golint error 

soniasingla:


natalieparellano:
Since we're not actually trying to produce HTML here I wonder if it's better to just ignore the warning.

soniasingla:
> Since we're not actually trying to produce HTML here I wonder if it's better to just ignore the warning.

Should I add ``// ignore`` comment? @natalieparellano

natalieparellano:
Apologies for the slow reply! I think for now that is the best option.

natalieparellano (approved):

natalieparellano:
I was trying to see if there's a more granular way to suppress the warning (like `// #nosec G204`), but for now I don't think so.

yaelharel (approved):

PR #110, opened 2021-05-05
Replaces test registry container with in-process instance

micahyoung:
### Overview
This re-works the registry helper to remove registry containers entirely and instead run and in-process `ggcr/registry` in a background go proc within the test process itself, still allowing tests to write blobs and manifests which are stored in memory. Like before this still uses random ports and HTTP Basic Auth for writing but allows unauthenticated read-only requests.

The motivation was to simplify the registry model to avoid the recent  Docker Desktop networking [regressions](https://github.com/docker/for-mac/issues?q=is%3Aissue+is%3Aopen+network+%22host%22+), particularly for `--network=host`. This also makes the tests roughly 15% faster.

This branch is working for lifecycle too (see [draft PR](https://github.com/buildpacks/lifecycle/pull/607)) on GHA, even with complex tests like analyzer acceptance which use `--network=host`. But like `lifecycle/main`, though, that PR branch won't pass on a MacOS Docker Desktop 3.3.x without an `insecure-registries` entry (though with this change, the entry can now be more broad, like `192.168.0.0/16` instead of `192.168.1.100/32`).

### From commit
- Simplifies and speeds up helper registry initialization
- Uses ggcr/registry (runs in test process, in-memory storage)
- Change remote.Delete test since ggcr/registry does not delete :latest tag when underlying image digest is deleted


remote/remote_test.go
```diff
@@ -1470,10 +1470,15 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertNil(t, err)
+
+				repoID := identifier.String()
+
 				img, err = remote.NewImage(
-					repoName,
+					repoID,
 					authn.DefaultKeychain,
-					remote.FromBaseImage(repoName),
+					remote.FromBaseImage(repoID),
```
micahyoung, line 1481: Without this change, this test was failing saying that after deletion, the image was still present. My assumption is that `ggcr/registry` treats deletions differently that `docker-distribution`. 

I think `remote.Image.Delete()` deletes by the identifier but `remote.Image.Found()` checks by tag and `docker-distribution` cleans up "hanging" tags whereas `ggcr/registry` does not.

That's just my assumption, I'd like to hear other ideas or suggestions.

remote/remote_test.go
```diff
@@ -1491,7 +1496,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				h.AssertNil(t, err)
 
 				h.AssertEq(t, img.Found(), false)
-				h.AssertError(t, img.Delete(), "MANIFEST_UNKNOWN")
+				h.AssertError(t, img.Delete(), "NAME_UNKNOWN")
```
jabrown85, line 1499: These are explicitly different errors [here](https://github.com/google/go-containerregistry/blob/83f4080544c8578ec02b937c2a0a4dc07198916d/pkg/registry/manifest.go#L96-L111). The `NAME_UNKNOWN` is if the entire repo doesn't exist, whereas `MANIFEST_UNKNOWN` means a specific manifest under that repo is missing.

It seems weird the entire repo is gone after we delete...

tools/Dockerfile.windows
```diff
@@ -0,0 +1,36 @@
+ARG image_tag
```
jabrown85, line 1: Unrelated?

Makefile
```diff
@@ -2,6 +2,8 @@
 GOCMD?=go
 GO_VERSION=$(shell go list -m -f "{{.GoVersion}}")
 PACKAGE_BASE=github.com/buildpacks/imgutil
+WINDOWS_COMPILATION_IMAGE?=golang:1.15-windowsservercore-1809
```
jabrown85, line 5: Unrelated?

Makefile
```diff
@@ -40,3 +42,14 @@ endif
 
 test: layer/bcdhive_generated.go format lint
 	$(GOCMD) test -parallel=1 -count=1 -v ./...
+
+# Ensure workdir is clean and build image from .git
+docker-build-source-image-windows:
```
jabrown85, line 47: Unrelated?

micahyoung: Oops, totally forgot to to cover this. I'll separate this out into a different PR.

remote/remote_test.go
micahyoung: Ah, interesting. I assumed this was just something different about `ggcr/registry`'s cleanup mechanisms but perhaps it's a bug in `Delete` that `docker-distribution` was covering up. I'll see if I can set up some test scenarios to figure out what's going on.

Makefile
micahyoung: Reverted for now, I'll bring it back separately.

remote/remote_test.go
micahyoung: This ends up being consistent with `gcr.io` behavior. Using a ref like,`gcr.io/dotnet-build-gcp/notarepo:latest`, where `dotnet-build-gcp` is a project but `notarepo:latest` doesn't exist and doing something like:
```go
image, err := remote.NewImage("gcr.io/dotnet-build-gcp/notarepo:latest", authn.DefaultKeychain)
image.Delete() // returns `NAME_UNKNOWN`
```

Output:
```
2021/05/07 07:25:29 DELETE https://gcr.io/v2/dotnet-build-gcp/neverthere/manifests/sha256:03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10: NAME_UNKNOWN: Failed to compute blob liveness for manifest: 'sha256:03cbce912ef1a8a658f73c660ab9c539d67188622f00b15c4f15b89b884f0e10'
exit status 1
```

testhelpers/docker_registry.go
```diff
@@ -64,143 +59,82 @@ func NewDockerRegistry(ops ...RegistryOption) *DockerRegistry {
 	return registry
 }
 
-func (r *DockerRegistry) Start(t *testing.T) {
-	t.Helper()
-
-	r.Host = DockerHostname(t)
-
-	t.Logf("run registry on %s", r.Host)
+func BasicAuth(handler http.Handler, username, password, realm string) http.Handler {
```
natalieparellano, line 62: Do this and `ReadOnly` need to be public?

testhelpers/docker_registry.go
```diff
@@ -64,143 +59,82 @@ func NewDockerRegistry(ops ...RegistryOption) *DockerRegistry {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		user, pass, ok := r.BasicAuth()
 
-	PullIfMissing(t, DockerCli(t), registryImageName)
+		if !ok || user != username || pass != password {
+			w.Header().Set("WWW-Authenticate", `Basic realm="`+realm+`"`)
+			w.WriteHeader(401)
+			_, _ = w.Write([]byte("Unauthorised.\n"))
```
natalieparellano, line 69:
```suggestion
			_, _ = w.Write([]byte("Unauthorized.\n"))
```

natalieparellano: LOL

testhelpers/docker_registry.go
```diff
@@ -64,143 +59,82 @@ func NewDockerRegistry(ops ...RegistryOption) *DockerRegistry {
...
-		if info.OSType == "windows" {
-			containerUser = "ContainerAdministrator" //required for volume permissions
-			storageBindPath = "c:/registry-storage"
-		}
+	r.Host = DockerHostname(t)
 
-		volumeBinds = append(volumeBinds, fmt.Sprintf("%s:%s", r.volumeName, storageBindPath))
+	// create registry handler, if not re-using a shared one
+	if r.regHandler == nil {
+		// change to os.Stderr for verbose output
```
natalieparellano, line 96: Is this a thing? Do we give this knob to the consumer?

testhelpers/docker_registry.go
```diff
@@ -4,42 +4,37 @@ import (
...
-var registryImageName = "micahyoung/registry:latest"
-
 type RegistryOption func(registry *DockerRegistry)
 
-//WithSharedStorageVolume allows two instances to share the same data volume.
+//WithSharedHandler allows two instances to share the same data by re-using the registry handler.
 //Use an authenticated registry to write to a read-only unauthenticated registry.
-//Volumes that don't exist will be created, then removed on Stop().
-func WithSharedStorageVolume(volumeName string) RegistryOption {
+func WithSharedHandler(handler http.Handler) RegistryOption {
```
natalieparellano, line 35: While we're here I wonder if it would make sense to rename `WithAuth` to `WithDockerConfig`. For some reason I find this easier to understand but maybe it's just me!

natalieparellano: Actually I realized that then this would read `ih.NewDockerRegistry(ih.WithDockerConfig(dockerConfigDir), ih.WithSharedHandler(sharedRegHandler))` and the extra docker might be redundant. I don't know... I'm just speaking aloud here as I'm trying to remember how this code works.

natalieparellano (approved):

natalieparellano:
Added a couple of comments but overall this looks great! I'm very excited about these improvements!

@jabrown85 do you want to take another look?

testhelpers/docker_registry.go
micahyoung: I was on the fence but since there's no auth built-into `ggcr/registry`, it could be useful to expose these. Also because this `imgutil` helper will be external to `lifecycle` and would need to be changed and bump to add functionality, leaving these public could make it easier to make a workaround while a change is being contributed into `imgutil`.

That is just hypothetical though, so if we'd rather make them private, that's fine with me.

testhelpers/docker_registry.go
micahyoung: `regHandler` would only have been set by `WithSharedHandler`: https://github.com/buildpacks/imgutil/pull/110/files/f4c144df5e3abddb910e47e17b95b64bf82c7af4#diff-66c9268014342d51b1c5446732d8eb82958f0f8c1c04e7199033662cabd8a35cR37

testhelpers/docker_registry.go
micahyoung: I feel like `WithDockerConfig` might miss that the new randomly generated username and password are also created (which later are used to branch on whether to be read-only or not). Perhaps `WithAuthAndConfig` or something?

testhelpers/docker_registry.go
natalieparellano: Makes sense! I'm happy to leave it as is.

jabrown85 (approved):

PR #111, opened 2021-05-05
Improving retry logic in remote to handle more cases

jjbustamante:
### References

- This Pull Request is created to fix the issue [#1069 in Pack repository](https://github.com/buildpacks/pack/issues/1069)

### Context

A previous [Pull Request #96](https://github.com/buildpacks/imgutil/pull/96) was merged recently to add some retry logic when an EOF is thrown. The logic was extended to handle the cases in the ticket  [#1069 in Pack](https://github.com/buildpacks/pack/issues/1069).  

### Description

- The PR adds some unit tests to verify the behavior of the retry logic implemented in remote.NewV1Image method
- A little bug was fixed, when there is not problem during the communication with the docker registry, the current was been retried two more times (a break was added to avoid this)


remote/remote.go
```diff
@@ -207,11 +207,24 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
 			}
 			return nil, fmt.Errorf("connect to repo store '%s': %s", repoName, err.Error())
 		}
+		break
```
natalieparellano, line 210: Thanks for catching this. Uggh 300ms extra for every successful image fetch... this might be worthy of a patch WDYT @dfreilich @jabrown85 @jromero ?

jromero: I'm pro patch if the burden is minimal.

remote/remote.go
```diff
@@ -207,11 +207,24 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
...
 
+func isRetryable(err error) bool {
+	if err == io.EOF {
+		return true
+	}
+	if transportErr, ok := err.(*transport.Error); ok {
+		switch transportErr.StatusCode {
+		case http.StatusNotFound, http.StatusUnauthorized:
+			return true
+		}
```
yaelharel, line 223:
```suggestion
		if transportErr.StatusCode ==  http.StatusNotFound || transportErr.StatusCode == http.StatusUnauthorized {
			return true
		}
```

I think that we usually use `switch-case` when for each result, we should act differently.
You can also have only one "big" if condition for all of this `transportErr`, but I'm not sure whether it'll be more readable. I don't have strong opinion on whether it should be one big if or two nested ifs.

yaelharel: I see that in the `if` condition on line 199, we also check that `len(transportErr.Errors) > 0`.
Do we really need it? And if we do, do we need to add it to the condition on line 219?

remote/remote.go
```diff
@@ -207,11 +207,24 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
...
 	}
-
 	return image, nil
 }
 
+func isRetryable(err error) bool {
+	if err == io.EOF {
+		return true
+	}
+	if transportErr, ok := err.(*transport.Error); ok {
```
yaelharel, line 219: I see that in the if condition on line 199, we also check that len(transportErr.Errors) > 0.
Do we really need it? And if we do, do we need to add it to the condition on line 219?

remote/remote.go
jjbustamante: Honestly looking both options if statement looks for readable to me

aemengo: Sorry if this was already covered, but this has been confirmed that we're actually seeing `401` and `404` and then _retrying_ the requests makes these go away?

I'm asking because `40x` errors are usually caused by the client, not by the server or by network interference.

jjbustamante: Well, the only way I managed to reproduce the error mentioned by @jabrown85  in the [ticket 1069](https://github.com/buildpacks/pack/issues/1069) was having the newV1Image method returning an empty image, and that happens when we get 404, 401 or an error with the message "no child with platform", I am not very happy to retry with 40x but otherwise not sure how to address the issue

aemengo: I see. Well, works for me 🙂

jjbustamante: @yaelharel I just added the suggestion about the if statement

natalieparellano: @micahyoung do you have any thoughts here about the 40x errors?

remote/remote_test.go
```diff
@@ -20,6 +26,15 @@ import (
...
 
 var dockerRegistry, readonlyDockerRegistry *h.DockerRegistry
+var count int
+
+type RemoteTestCase struct {
+	name        string
+	repo        string
+	statusCode  int
+	failedCount int
+	expected    int
```
natalieparellano, line 36:
```suggestion
	expectedTries    int
```

remote/remote_test.go
```diff
@@ -1496,3 +1510,61 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			img, _ := random.Image(1024, 1)
+			configName, _ := img.ConfigName()
+			server := createServer(img, configName, tc)
+			u, err := url.Parse(server.URL)
+			if err != nil {
+				t.Fatalf("url.Parse(%v) = %v", server.URL, err)
+			}
```
natalieparellano, line 1529:
```suggestion
			h.AssertNil(t, err)
```

remote/remote_test.go
```diff
@@ -1496,3 +1510,61 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			defer server.Close()
+			var repoName = u.Hostname() + ":" + u.Port() + "/" + tc.repo
+
+			// exercise the subject
+			count = 0
+			_, err = remote.NewImage(repoName, authn.DefaultKeychain, remote.WithPreviousImage(repoName))
+			if err != nil {
+				t.Fatalf("Error: %v, running remoteNewImage", err)
+			}
```
natalieparellano, line 1538:
```suggestion
			h.AssertNil(t, err)
```

natalieparellano: For consistency with existing patterns in this file. You could argue however that the way you have it produces a more descriptive error message...

remote/remote_test.go
```diff
@@ -1496,3 +1510,61 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+	manifestPath := fmt.Sprintf("/v2/%s/manifests/latest", tc.repo)
+	configPath := fmt.Sprintf("/v2/%s/blobs/%s", tc.repo, configName)
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var err error
+		switch r.URL.Path {
+		case "/v2/":
+			w.WriteHeader(http.StatusOK)
+		case configPath:
+			c, _ := img.RawConfigFile()
+			_, err = w.Write(c)
```
natalieparellano, line 1556: These seem not to be necessary for the tests to pass.

remote/remote_test.go
```diff
@@ -1496,3 +1510,61 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 }
+
+func TestRemoteRetryLogic(t *testing.T) {
```
natalieparellano, line 1514: This is awesome! My one nit is if you could please convert this to using `spec` notation and nest it under this `when`: https://github.com/buildpacks/imgutil/blob/77564c120f1445ad6c9c0ca3dfd8de99be016279/remote/remote_test.go#L63

That way it's easier to find the test and debug with consistent notation (like `it.Focus`).

remote/remote_test.go
natalieparellano: (both `"/v2/"` and `configPath`)

remote/remote_test.go
natalieparellano: IDK, this seems clearer.

remote/remote_test.go
```diff
@@ -20,6 +26,15 @@ import (
 )
 
 var dockerRegistry, readonlyDockerRegistry *h.DockerRegistry
+var count int
```
natalieparellano, line 29: I can see why you did it this way, but is there any way to avoid making this a global variable?

remote/remote.go
```diff
@@ -207,11 +207,24 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
...
+		if transportErr.StatusCode == http.StatusNotFound ||
+			transportErr.StatusCode == http.StatusUnauthorized {
+			return true
+		}
+	}
```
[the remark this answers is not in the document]
yaelharel: It's not the exact same logic since in the link you posted we also check that `len(transportErr.Errors) > 0`

remote/remote_test.go
jjbustamante: Hi @natalieparellano  you mean do not using the table driven approach, right?

remote/remote_test.go
jjbustamante: Any suggestion will be great!, let me try to find some other way to do it

remote/remote_test.go
jjbustamante: "/v2" is required, because it's the first call made by the client, but other one, the configPath I think it's not, let me run the tests without it and I can removes it if it is not necessary, I think I added because the first attempt was made using the remote.fromBaseImage method and that one was actually using the configPath, but I had an issue with the OS in the random image not set up, and I change to test it with remote.WithPreviosImage

remote/remote_test.go
jjbustamante: What about something like this:

`
if err != nil {
	// some code ..
	if transportErr, ok := err.(*transport.Error); ok && len(transportErr.Errors) > 0 && is40x(err) {
		return emptyImage(platform)
	}
	// more code ...
}

func isRetryable(err error) bool {
	if err == io.EOF {
		return true
	}
	return is40x(err)
}

func is40x(err error) bool {
	if transportErr, ok := err.(*transport.Error); ok {
		if transportErr.StatusCode == http.StatusNotFound ||
			transportErr.StatusCode == http.StatusUnauthorized {
			return true
		}
	}
	return false
}
`

remote/remote.go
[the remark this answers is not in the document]
jjbustamante: What about something like this:

```
if err != nil {
	// some code ..
	if transportErr, ok := err.(*transport.Error); ok && len(transportErr.Errors) > 0 && is40x(err) {
		return emptyImage(platform)
	}
	// more code ...
}

func isRetryable(err error) bool {
	if err == io.EOF {
		return true
	}
	return is40x(err)
}

func is40x(err error) bool {
	if transportErr, ok := err.(*transport.Error); ok {
		if transportErr.StatusCode == http.StatusNotFound ||
			transportErr.StatusCode == http.StatusUnauthorized {
			return true
		}
	}
	return false
}
```

remote/remote_test.go
jjbustamante: Applied

remote/remote_test.go
jjbustamante: Applied

remote/remote_test.go
jjbustamante: Applied

remote/remote_test.go
jjbustamante: @natalieparellano you were right, they are not needed for the test to pass, I just removed those lines

remote/remote.go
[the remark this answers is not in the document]
yaelharel: I don't like the idea that in the line:
```
if transportErr, ok := err.(*transport.Error); ok && len(transportErr.Errors) > 0 && is40x(err) {
```
you're running the following code twice:
```
if transportErr, ok := err.(*transport.Error); ok {
```

Maybe you can pass the `transportErr` to the `is40x` function and then it'll contain only the nested `if` in it. Something like:
```
func is40x(transportErr transport.Error) bool {
    return transportErr.StatusCode == http.StatusNotFound || transportErr.StatusCode == http.StatusUnauthorized
}
```

remote/remote_test.go
yaelharel: I think what Natalie means is to keep the convention that we use in our unit tests:

As you can see in line 48 in your code, we have the `TestRemote` function. At the end of this function we call:
`spec.Run(t, "Image", testImage, spec.Sequential(), spec.Report(report.Terminal{}))`.
Then in the `testImage` function, we're testing (most of) the public functions of `remote.go`. Each one under a different `when` section.
Since your changes in `remote.go` are part of the `NewImage` flow, it would be great if you could put your tests under the `NewImage` section (`when("#NewImage", func()` in line 77), while using nested `when` and `it` for each `testCase`.

@natalieparellano, please let me know if I misunderstood what you meant.

remote/remote.go
micahyoung: > the only way I managed to reproduce the error ... was having the newV1Image method returning an empty image

If you're able to wait until https://github.com/buildpacks/imgutil/pull/110 is merged (or willing to retrofit), I found it's pretty easy to wrap a `ggcr/registry` with custom request/response logic like I do in my PR to get read-only behavior:
https://github.com/buildpacks/imgutil/blob/f27928a6bb8e4ddf6694f55baa3d7413efba6c6b/testhelpers/docker_registry.go#L77-L87

So you could potentially make a new "flaky" handler like:
```go
func DropFirstRequest(handler http.Handler) http.Handler {
        firstRequest := true
		if firstRequest {
                         firstRequest = false
			return
		}

		handler.ServeHTTP(w, r)
	})
}
``` 

From which you could make a new "flaky" read-only registry: 
```go
dropFirstRequestHandler := DropFirstRequest(sharedRegistryHandler)
flaklyDockerRegistry = h.NewDockerRegistry(h.WithSharedHandler(dropFirstRequestHandler))
flaklyDockerRegistry.Start(t)
defer flaklyDockerRegistry.Stop(t)
```

remote/remote.go
jjbustamante: Included in on [PR 117](https://github.com/buildpacks/imgutil/pull/117)

remote/remote_test.go
yaelharel: Maybe when you make [these changes](https://github.com/buildpacks/imgutil/pull/111#discussion_r627863324), it'll be easier to move this global variable.

natalieparellano: I am not sure if this would work, but perhaps the server could hold on to a map of endpoints -> count of times accessed. Then you could have a separate endpoint to retrieve the map? IDK

jjbustamante: I created a new data structure wrapping the http server and saving the number of methods calls internally, in that way I managed to remove the global variable

remote/remote.go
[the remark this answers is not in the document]
yaelharel: When you call `transportErr, ok := err.(*transport.Error)`, this is a type assertion.
As you can read [here](https://www.sohamkamani.com/golang/type-assertions-vs-type-conversions/):
"`ok` is a boolean value which is `true` if our assertion is correct, and false otherwise."

remote/remote_test.go
```diff
@@ -501,6 +512,59 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+						_, err := remote.NewImage(repoName, authn.DefaultKeychain, remote.WithPreviousImage(repoName))
+						h.AssertNil(t, err)
+						if tc.expectedTries != mockServer.ActualCount() {
+							t.Fatalf("Expected number of invocations %d different than actual value %d",
+								tc.expectedTries, mockServer.ActualCount())
+						}
+					})
+				})
+			}
+		})
```
yaelharel, line 567: I think it's better to have a separate nested `when` for each test case so it'll be easier to investigate failures in the future (the output will indicate the specific use case that failed).
In order to prevent code duplication, you can create some helper functions.

remote/remote.go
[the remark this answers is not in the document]
yaelharel: Marking this conversation as "unresolved", only so you'll be able to read my previous comment (otherwise, Github will hide it). Please feel free to "resolve" this conversation again.

remote/remote.go
```diff
@@ -193,14 +193,13 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
 		time.Sleep(100 * time.Duration(i) * time.Millisecond) // wait if retrying
 		image, err = remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport), remote.WithPlatform(v1Platform))
 		if err != nil {
-			if err == io.EOF && i != maxRetries {
+			transportErr, ok := err.(*transport.Error)
+			isValid40x := ok && is40x(transportErr)
```
yaelharel, line 197: Please see https://github.com/buildpacks/imgutil/pull/111#discussion_r631035070.

remote/remote_test.go
jjbustamante: I refactored the code and put each test case on a separate `when`,  `it` statement. I didn't find a good way to prevent code duplication without affecting the readability, I kept each test case with its own code.

remote/remote_test.go
```diff
@@ -21,6 +24,15 @@ import (
...
 var dockerRegistry, readonlyDockerRegistry *h.DockerRegistry
 
+type RemoteTestCase struct {
+	when          string
+	it            string
+	repo          string
+	statusCode    int
+	failedCount   int
+	expectedTries int
+}
```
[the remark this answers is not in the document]
jjbustamante: After refactoring the test cases, I removed this code.

remote/remote_test.go
```diff
@@ -501,6 +503,77 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
 			})
 		})
+
+		when("#RetryLogic", func() {
+			var server *httptest.Server
+			var repo string
+
+			when("Manifest API in registry returns status code 200", func() {
+				repo = "org/do-not-retry"
+				mockServer := h.NewMockServer(repo, http.StatusOK, 0)
```
yaelharel, line 513: Why aren't these lines under the `it.Before`?

yaelharel: If you'll move it to be under `it.Before`, you won't need line 509 but `mockServer` will need to be defined there instead.

remote/remote_test.go
```diff
@@ -501,6 +503,77 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				repo = "org/retry-not-found"
+				mockServer := h.NewMockServer(repo, http.StatusNotFound, 2)
+
+				it.Before(func() {
+					server = mockServer.Server()
+					u, err := url.Parse(server.URL)
+
+					h.AssertNil(t, err)
+
+					repoName = u.Hostname() + ":" + u.Port() + "/" + repo
```
yaelharel, line 543: I think you can create an helper function for this in order to prevent code duplication.

remote/remote_test.go
```diff
@@ -501,6 +503,77 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				})
+
+				it("Retry after status code 404", func() {
+					defer server.Close()
+					_, err := remote.NewImage(repoName, authn.DefaultKeychain, remote.WithPreviousImage(repoName))
+
+					h.AssertNil(t, err)
+					h.AssertEq(t, mockServer.ActualCount(), 3)
```
yaelharel, line 551: Same comment as https://github.com/buildpacks/imgutil/pull/111/files#r631156024.

remote/remote_test.go
jjbustamante: Yes, I just noticed that

jjbustamante: I managed to simplify the variable declaration and only the MockServer was required.

remote/remote_test.go
jjbustamante: I did, I put the function in `testhelpers.go` file.

remote/remote_test.go
jjbustamante: I put it in the same `remote_test.go` file because I need to access the remote package, is that ok?

remote/remote_test.go
```diff
@@ -1569,3 +1530,11 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 }
+
+func VerifyRetryLogic(t *testing.T, mockServer *h.MockServer, repoName string, expectedCount int) {
```
yaelharel, line 1534: It doesn't have to be public. You can change it to `verifyRetryLogic`.

testhelpers/testhelpers.go
```diff
@@ -409,3 +410,14 @@ func checkResponseError(r io.Reader) error {
 
 	return nil
 }
+
+func SetUpMockServer(t *testing.T, repo string, statusCode, failedCount int) (*MockServer, string) {
```
yaelharel, line 414: What do you think about moving this function to `mock_registry.go`?

jjbustamante: I thought about it at first, but then I was not sure, and because of this testhelpers are other methods reuse in other places I decided to put it here. But, I think it will not harm anyone if we move it to `mock_registry.go`

remote/remote_test.go
jjbustamante: Oh. Sorry, I didn't notice the difference between the capital letter or not during the method definition.  I just googled it, I agreed it should be not exported.

remote/remote_test.go
```diff
@@ -501,6 +501,40 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				})
 			})
 		})
+
+		when("#RetryLogic", func() {
```
natalieparellano, line 505:
```suggestion
		when("retry logic", func() {
```
(for other scopes that have `#Thing`, `Thing` is an accessible function from the package)

remote/remote_test.go
```diff
@@ -501,6 +501,40 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			var mockServer *h.MockServer
+
+			when("Manifest API in registry returns status code 200", func() {
```
natalieparellano, line 508:
```suggestion
			when("manifest API in registry returns status code 200", func() {
```

testhelpers/mock_registry.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (m *MockServer) ActualCount() int {
+	return m.actualCount
+}
+
+func (m *MockServer) Server() *httptest.Server {
+	return m.server
+}
+
+func SetUpMockServer(t *testing.T, repo string, statusCode, failedCount int) (*MockServer, string) {
```
natalieparellano, line 59:
```suggestion
func SetUpMockServer(t *testing.T, repo string, initialCode, timesToFail int) (*MockServer, string) {
```

natalieparellano: (just a suggestion... this is how I understand it :))

remote/remote_test.go
```diff
@@ -1496,3 +1530,11 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 }
+
+func verifyRetryLogic(t *testing.T, mockServer *h.MockServer, repoName string, expectedCount int) {
+	defer mockServer.Server().Close()
```
natalieparellano, line 1535: This might make more sense in an `it.After`

remote/remote_test.go
```diff
@@ -1496,3 +1530,11 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 }
+
+func verifyRetryLogic(t *testing.T, mockServer *h.MockServer, repoName string, expectedCount int) {
```
natalieparellano, line 1534:
```suggestion
func assertExpectedRetries(t *testing.T, mockServer *h.MockServer, repoName string, expectedCount int) {
```

remote/remote_test.go
```diff
@@ -501,6 +501,40 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				it.Before(func() {
+					mockServer, repoName = h.SetUpMockServer(t, "org/do-not-retry", http.StatusOK, 0)
+				})
+
+				it("Do not retry after status code 200", func() {
+					verifyRetryLogic(t, mockServer, repoName, 1)
+				})
+			})
+
+			when("Manifest API in registry returns status code 404", func() {
```
natalieparellano, line 518:
```suggestion
			when("manifest API in registry returns status code 404", func() {
```

remote/remote_test.go
```diff
@@ -501,6 +501,40 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				it.Before(func() {
+					mockServer, repoName = h.SetUpMockServer(t, "org/retry-not-found", http.StatusNotFound, 2)
+				})
+
+				it("Retry after status code 404", func() {
```
natalieparellano, line 523:
```suggestion
				it("retries after status code 404", func() {
```

remote/remote_test.go
```diff
@@ -501,6 +501,40 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+					verifyRetryLogic(t, mockServer, repoName, 3)
+				})
+			})
+
+			when("Manifest API in registry returns status code 401", func() {
```
natalieparellano, line 528:
```suggestion
			when("manifest API in registry returns status code 401", func() {
```

remote/remote_test.go
```diff
@@ -501,6 +501,40 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				it.Before(func() {
+					mockServer, repoName = h.SetUpMockServer(t, "org/retry-unauthorized", http.StatusUnauthorized, 2)
+				})
+
+				it("Retry after status code 401", func() {
```
natalieparellano, line 533:
```suggestion
				it("retries after status code 401", func() {
```

remote/remote_test.go
jjbustamante: I agreed with all the other suggestions, but this one makes me think.  If I refactor the code for actually exercising the SUT (System Under Test) and moves the verification to the `it.After` it will look like this:
```
when("manifest API in registry returns status code 200", func() {
	it.Before(func() {
		mockServer, repoName = h.SetUpMockServer(t, "org/do-not-retry", http.StatusOK, 0)
	})

	it("does not retry after status code 200", func() {
		exerciseRetryLogic(t, mockServer, repoName)
	})

	it.After(func() {
		h.AssertEq(t, mockServer.ActualCount(), 1)
	})
})

func exerciseRetryLogic(t *testing.T, mockServer *h.MockServer, repoName string) {
	defer mockServer.Server().Close()
	_, err := remote.NewImage(repoName, authn.DefaultKeychain, remote.WithPreviousImage(repoName))
	h.AssertNil(t, err)
}
```

I will not need the Assertion method, but I can keep the exercise in a different method to avoid duplication. How does it look like? @natalieparellano

remote/remote.go
```diff
@@ -193,14 +193,13 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
 		time.Sleep(100 * time.Duration(i) * time.Millisecond) // wait if retrying
 		image, err = remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport), remote.WithPlatform(v1Platform))
 		if err != nil {
-			if err == io.EOF && i != maxRetries {
+			transportErr, ok := err.(*transport.Error)
```
natalieparellano, line 196: For some reason I find this code a bit difficult to read. What do you think of the following?
```
if err != nil {
			if err == io.EOF && i != maxRetries {
				continue // retry
			}
			if transportErr, ok := err.(*transport.Error); ok && is40x(transportErr) {
				if i != maxRetries {
					continue // retry
				}
				if len(transportErr.Errors) > 0 { // will this always be true for 40x ?
					return emptyImage(platform)
				}
			}
			if strings.Contains(err.Error(), "no child with platform") {
				return emptyImage(platform)
			}
			return nil, fmt.Errorf("connect to repo store '%s': %s", repoName, err.Error())
		}
```

remote/remote.go
```diff
@@ -193,14 +193,13 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
...
+			isValid40x := ok && is40x(transportErr)
+			if err == io.EOF || isValid40x && i != maxRetries {
```
natalieparellano, line 198: I think this has the subtle bug that @yaelharel found here: https://github.com/buildpacks/imgutil/pull/99/files (fixed in the above suggestion)

testhelpers/mock_registry.go
```diff
@@ -0,0 +1,51 @@
...
+	img, _ := random.Image(1024, 1)
+	m.server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var err error
+		if r.URL.Path == manifestPath {
+			m.actualCount++
+			if m.actualCount <= m.failedCount {
+				w.WriteHeader(m.statusCode)
+			} else {
+				mm, _ := img.RawManifest()
+				_, err = w.Write(mm)
```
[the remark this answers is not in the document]
natalieparellano: @muse-dev ignore

remote/remote_test.go
natalieparellano: Oh sorry, I meant just `defer mockServer.Server().Close()` in the `After`

natalieparellano:
Made a few more suggestions for readability. There is one subtle bug to fix (see https://github.com/buildpacks/imgutil/pull/111#discussion_r632793262 ) but overall this is looking good!

remote/remote.go
yaelharel: You're so right!
We can move `i != maxRetries` to the beginning of the condition but I think that your suggestion above is better.

yaelharel: @jjbustamante, just to make it clear, if we got an error after `maxRetries`, we want to return an error that is defined in the `for` loop and not get to line 211.

jjbustamante: Yeap, got it! Thanks!

natalieparellano:
I wonder if https://github.com/google/go-containerregistry/pull/1019 is relevant to this PR.

remote/remote_test.go
```diff
@@ -502,36 +503,48 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
-			when("Manifest API in registry returns status code 200", func() {
+			when("manifest API in registry returns status code 200", func() {
 				it.Before(func() {
 					mockServer, repoName = h.SetUpMockServer(t, "org/do-not-retry", http.StatusOK, 0)
 				})
 
-				it("Do not retry after status code 200", func() {
-					verifyRetryLogic(t, mockServer, repoName, 1)
+				it.After(func() {
+					defer mockServer.Server().Close()
```
natalieparellano, line 515:
```suggestion
					mockServer.Server().Close()
```

remote/remote_test.go
```diff
@@ -502,36 +503,48 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				})
+
+				it("does not retry after status code 200", func() {
+					assertExpectedRetries(t, mockServer, repoName, 1)
```
natalieparellano, line 519:
```suggestion
					assertExpectedTries(t, mockServer, repoName, 1)
```

natalieparellano: (I'm sorry, I know I originally suggested `assertExpectedRetries` - but the current phrasing makes me think that we still retry once even on a successful request!)

remote/remote_test.go
```diff
@@ -502,36 +503,48 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
-			when("Manifest API in registry returns status code 404", func() {
+			when("manifest API in registry returns status code 404", func() {
 				it.Before(func() {
 					mockServer, repoName = h.SetUpMockServer(t, "org/retry-not-found", http.StatusNotFound, 2)
 				})
 
-				it("Retry after status code 404", func() {
-					verifyRetryLogic(t, mockServer, repoName, 3)
+				it.After(func() {
+					defer mockServer.Server().Close()
```
natalieparellano, line 529:
```suggestion
					mockServer.Server().Close()
```

remote/remote_test.go
```diff
@@ -502,36 +503,48 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
-			when("Manifest API in registry returns status code 401", func() {
+			when("manifest API in registry returns status code 401", func() {
 				it.Before(func() {
 					mockServer, repoName = h.SetUpMockServer(t, "org/retry-unauthorized", http.StatusUnauthorized, 2)
 				})
 
-				it("Retry after status code 401", func() {
-					verifyRetryLogic(t, mockServer, repoName, 3)
+				it.After(func() {
+					defer mockServer.Server().Close()
```
natalieparellano, line 543:
```suggestion
					mockServer.Server().Close()
```

natalieparellano (approved):

natalieparellano:
From my perspective this looks good. I left a couple of nits :)

I do think it would be good to know what is the overlap between what this PR is trying to accomplish and https://github.com/google/go-containerregistry/pull/1019. Could we remove some of the logic from our side?

remote/remote_test.go
```diff
@@ -1543,7 +1543,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 	})
 }
 
-func assertExpectedRetries(t *testing.T, mockServer *h.MockServer, repoName string, expectedCount int) {
+func assertExpectedTries(t *testing.T, mockServer *h.MockServer, repoName string, expectedCount int) {
 	_, err := remote.NewImage(repoName, authn.DefaultKeychain, remote.WithPreviousImage(repoName))
 
 	h.AssertNil(t, err)
```
yaelharel, line 1549: Please add an empty line at the end of the file.

remote/remote.go
```diff
@@ -194,11 +194,13 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
 		image, err = remote.Image(ref, remote.WithAuth(auth), remote.WithTransport(http.DefaultTransport), remote.WithPlatform(v1Platform))
 		if err != nil {
 			if err == io.EOF && i != maxRetries {
```
yaelharel, line 196: It doesn't relate to your commit but I've just realized that `maxRetries` is set to `2` and the total number of retries is `3` because we start from `0`.
Maybe it's better to set `maxRetries` to 3 and change the `for loop` to be:
```
for i := 0; i < maxRetries; i++ {
```
I think it'll be clearer for the reader.

yaelharel (approved):
Looks good to me!

jjbustamante:
> From my perspective this looks good. I left a couple of nits :)
> 
> I do think it would be good to know what is the overlap between what this PR is trying to accomplish and [google/go-containerregistry#1019](https://github.com/google/go-containerregistry/pull/1019). Could we remove some of the logic from our side?

I think the logic implemented in that PR could affect the EOF logic implemented on our side.

```
if err == io.EOF &amp;&amp; i != maxRetries {				
     continue // retry			
}
```

We can remove it, but I think we can do that later, the worst case I see is an EOF is thrown and it will be handled for us, so this retry logic will not be executed, it doesn't harm anyone if we keep it. That's my opinion

remote/remote.go
```diff
@@ -194,11 +194,13 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
...
+				continue // retry
 			}
-			if transportErr, ok := err.(*transport.Error); ok && len(transportErr.Errors) > 0 {
-				switch transportErr.StatusCode {
-				case http.StatusNotFound, http.StatusUnauthorized:
+			if transportErr, ok := err.(*transport.Error); ok && is40x(transportErr) {
+				if i != maxRetries {
+					continue // retry
+				}
+				if len(transportErr.Errors) > 0 {
```
jabrown85, line 203: I am not familiar with the `transport.Error` - what is the purpose of the guard `len(transportErr.Errors) > 0`? I see it existed before, but was it guarding a failure of `transportErr.StatusCode`? The history of the files makes it look like we used to have this guard to protect `transportErr.Errors[0].Code` - but since we aren't doing that anymore I think we could remove this check entirely.

remote/remote.go
```diff
@@ -194,11 +194,13 @@ func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platf
...
 		if err != nil {
 			if err == io.EOF && i != maxRetries {
-				continue // retry if EOF
+				continue // retry
 			}
-			if transportErr, ok := err.(*transport.Error); ok && len(transportErr.Errors) > 0 {
-				switch transportErr.StatusCode {
-				case http.StatusNotFound, http.StatusUnauthorized:
+			if transportErr, ok := err.(*transport.Error); ok && is40x(transportErr) {
+				if i != maxRetries {
```
jabrown85, line 200: What is the cost of retrying? If an image is truly not on a registry, what kind of impact are we introducing to that situation?

remote/remote_test.go
```diff
@@ -501,6 +502,52 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				it.After(func() {
+					mockServer.Server().Close()
+				})
+
+				it("retries after status code 404", func() {
+					assertExpectedTries(t, mockServer, repoName, 3)
+				})
+			})
+
+			when("manifest API in registry returns status code 401", func() {
```
jabrown85, line 537: It would be nice to have a test where we `fail` `succeed` to show that it works and assert we call twice but actually succeed.

jabrown85 (commented):
Minor questions/nits

remote/remote.go
jjbustamante: If I am not wrong @natalieparellano estimated that the following [fix](https://github.com/buildpacks/imgutil/pull/117)    was saving us 300 milliseconds on each retry, we are retrying 3 times I believe the impact is about 900 milliseconds overhead if the image actually doesn't exist in the registry

jabrown85: Are we concerned about that? I feel like I want to be able to send something in like `WithExpectedToExist()` 😄  so that the consumer of this function can add on the behavior. In the case of an app image in analyzer, for instance, I wouldn't want to retry on 404s since it is likely this is a first build situation. But for fetching a run image, I would expect it to exist.

remote/remote_test.go
jjbustamante: Wait, @jabrown85  the current tests is actually doing that. 

`mockServer, repoName = h.SetUpMockServer(t, "org/retry-unauthorized", http.StatusUnauthorized, 2)` the mock server is initialized to failed 2 times, on the third invocation it will respond with a good response.

Then

`assertExpectedTries(t, mockServer, repoName, 3)` is verifying the remote path was called 3 times and inside that verification, I am checking the method was executed succeded

```
func assertExpectedTries(t *testing.T, mockServer *h.MockServer, repoName string, expectedCount int) {
	_, err := remote.NewImage(repoName, authn.DefaultKeychain, remote.WithPreviousImage(repoName))
	h.AssertNil(t, err). // Method was succeded 
	h.AssertEq(t, mockServer.ActualCount(), expectedCount)
}
```

Maybe I can save how many times the bad response was returned and assert against that value? also, that number can be calculated based on the failed times configuration and the total number of invocations too

jabrown85: ah, ok. I didn't read that the next time would be successful. I think you are good to go

natalieparellano:
@jjbustamante do you mind if I close this PR? IIRC from Jesse's comment [here](https://github.com/buildpacks/imgutil/pull/111#discussion_r639957271) we wanted to step back and re-evaluate the approach to avoid adding retry seconds for images that we expect might not exist.

jjbustamante:
> @jjbustamante do you mind if I close this PR? IIRC from Jesse's comment [here](https://github.com/buildpacks/imgutil/pull/111#discussion_r639957271) we wanted to step back and re-evaluate the approach to avoid adding retry seconds for images that we expect might not exist.

No problem at all

PR #113, opened 2021-05-06
Update some errors of bugsabsh

soniasingla:

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -722,7 +722,10 @@ func untar(r io.Reader, dest string) error {
 			if err != nil {
 				return err
 			}
-			if _, err := io.Copy(fh, tr); err != nil {
+			if _, err := io.CopyN(fh, tr, 1024); err != nil {
+				if err == io.EOF {
+					break
+				}
 				fh.Close()
 				return err
 			}


local/local.go
```diff
@@ -722,7 +722,7 @@ func untar(r io.Reader, dest string) error {
 			if err != nil {
 				return err
 			}
-			if _, err := io.Copy(fh, tr); err != nil {
+			if _, err := io.CopyN(fh, tr); err != nil {
```
yaelharel, line 725: I think that `io.CopyN` gets 3 variables: https://golang.org/pkg/io/#CopyN
You can read more about the "Potential DoS vulnerability via decompression bomb" we got from Gosec and how to solve it [here](https://stackoverflow.com/questions/67327323/g110-potential-dos-vulnerability-via-decompression-bomb-gosec).

soniasingla: Done @yaelharel thanks

local/local.go
```diff
@@ -722,7 +722,11 @@ func untar(r io.Reader, dest string) error {
 			if err != nil {
 				return err
 			}
-			if _, err := io.Copy(fh, tr); err != nil {
+			if _, err := io.CopyN(fh, tr, 1024);
```
yaelharel, line 725: As you can see [here](https://github.com/buildpacks/imgutil/pull/113/checks?check_run_id=2570564280) and [here](https://github.com/buildpacks/imgutil/pull/113/checks?check_run_id=2570564240), the tests are failing because of the following typo:
```suggestion
			if _, err := io.CopyN(fh, tr, 1024) {
```

Note: You can get to the failing tests by clicking on "Details" next to the failing tests at the bottom of the [PR's conversation tab](https://github.com/buildpacks/imgutil/pull/113).

local/local.go
```diff
@@ -722,7 +722,15 @@ func untar(r io.Reader, dest string) error {
 			if err != nil {
 				return err
 			}
-			if _, err := io.Copy(fh, tr); err != nil {
+<<<<<<< HEAD
+			if _, err := io.CopyN(fh, tr); err != nil {
+=======
+			if _, err := io.CopyN(fh, tr, 1024) {
+				if err != nil {
```
natalieparellano, line 729: Could we move this to the previous line?

We'll also need to resolve the merge conflict :)

soniasingla: Done

soniasingla: thanks

soniasingla:
@natalieparellano can you help me i did not get why test case is failing :/

natalieparellano:
@soniasingla thanks for updating the PR! I apologize, I didn't have a good grasp before on what `io.CopyN` is doing. In the current implementation, we are reading only 1024 bytes total - hence the unexpected EOF failures in the tests.

We could put the call to `io.CopyN` in a `for` loop as described [here](https://stackoverflow.com/a/67392303/3453671) and this would satisfy muse-dev (maybe) however it's not any better at preventing DOS attack vs. `io.Copy`, since we are still blindly copying the whole image. We would need to come up with some reasonable limit on the total number of bytes we expect to read in these scenarios.

I created an issue https://github.com/buildpacks/imgutil/issues/120 to capture this question so that we could discuss as a team and come up with a recommendation. If you have any thoughts here, we would love to hear them!

yaelharel:
I'm going to close this PR following https://github.com/buildpacks/imgutil/pull/113#issuecomment-844445287.
We'll discuss this change as part of https://github.com/buildpacks/imgutil/issues/120.

PR #123, opened 2021-06-03
Adding methods to check read/write access in Image

jjbustamante:

## Summary

This PR is adding a couple of methods in the Image object to verify if the client has read/write access in the registry, these methods will be used to fix the issue [572 - validate registry credentials](https://github.com/buildpacks/lifecycle/issues/572) in the [lifecycle](https://github.com/buildpacks/lifecycle) repository

## Documentation

    - [x] No

## Related
Resolves #572 (in lifecycle)


remote/remote.go
```diff
@@ -347,12 +347,36 @@ func (i *Image) Name() string {
...
+		if transportErr, ok := err.(*transport.Error); ok {
+			return transportErr.StatusCode != http.StatusUnauthorized &&
+				transportErr.StatusCode != http.StatusForbidden
+		}
+		return false
+	}
+	return true
+}
+
+func (i *Image) headManifest() (*v1.Descriptor, error) {
```
aemengo, line 374:
```suggestion
func (i *Image) found() (*v1.Descriptor, error) {
```

Love all this! Let's rename this to lowercase `found()`.
Having a similarly named **Uppercase/lowercase** methods is a (completely optional) golang idiom in these types of cases.

remote/remote_test.go
```diff
@@ -1500,4 +1500,82 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			it("returns false", func() {
+				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+				h.AssertEq(t, image.CheckReadAccess(), false)
+			})
+		})
+	})
+
+	when("#CheckReadWriteAccess", func() {
+		when("image exists in the registry and client has read access", func() {
```
natalieparellano, line 1544:
```suggestion
		when("image exists in the registry and client has read/write access", func() {
```

natalieparellano: (is this what you mean?)

remote/remote_test.go
```diff
@@ -1500,4 +1500,82 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+
+		when("image does not exist in the registry and client has read access", func() {
+			it("returns true", func() {
+				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+				h.AssertEq(t, image.CheckReadWriteAccess(), true)
+			})
+		})
+
+		when("image does not exist in the registry and client hasn't read access", func() {
```
natalieparellano, line 1565:
```suggestion
		when("image does not exist in the registry and client doesn't have read access", func() {
```

natalieparellano: Is there a way to make a "read only" test fixture?

remote/remote_test.go
```diff
@@ -1500,4 +1500,82 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			it("returns true", func() {
+				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+				h.AssertEq(t, image.CheckReadAccess(), true)
+			})
+		})
+
+		when("image does not exist in the registry and client hasn't read access", func() {
+			it.Before(func() {
+				os.Setenv("DOCKER_CONFIG", readonlyDockerRegistry.DockerDirectory)
```
natalieparellano, line 1528: Maybe this is just the variable name, but I find this confusing - shouldn't the client still have read access to the "read only" registry?

remote/remote_test.go
```diff
@@ -1500,4 +1500,82 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+
+		when("image does not exist in the registry and client has read access", func() {
+			it("returns true", func() {
+				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+				h.AssertEq(t, image.CheckReadAccess(), true)
+			})
+		})
+
+		when("image does not exist in the registry and client hasn't read access", func() {
```
natalieparellano, line 1526:
```suggestion
		when("image does not exist in the registry and client doesn't have read access", func() {
```

remote/remote_test.go
```diff
@@ -1500,4 +1500,82 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				h.AssertNil(t, origImage.Save())
+
+				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+
+				h.AssertEq(t, image.CheckReadWriteAccess(), true)
+			})
+		})
+
+		when("image does not exist in the registry and client has read access", func() {
```
natalieparellano, line 1557:
```suggestion
		when("image does not exist in the registry and client has read/write access", func() {
```

natalieparellano: Or maybe "only has read access" as mentioned here: https://github.com/buildpacks/imgutil/pull/123/files#r646862762

remote/remote_test.go
jjbustamante: When I switched to the read-only registry my auth.DefaultKeyChain is still pointing to dockerRegistry user/pass configuration. in this case, I am trying to access an image that doesn't exist in the registry with invalid credentials, that's why I do not have read access. 

I agreed, maybe the variable name is not appropriated, but it was already there, I didn't want to create another registry instance. Maybe I can point to the same registry with another variable name? could it make more clear?

remote/remote.go
jjbustamante: I addressed this suggestion in the following [commit](https://github.com/buildpacks/imgutil/pull/123/commits/d588cd706db0e32f55858fa00211c948d045210a)

remote/remote_test.go
jjbustamante: I addressed this suggestion in the following [commit](https://github.com/buildpacks/imgutil/pull/123/commits/d588cd706db0e32f55858fa00211c948d045210a)

remote/remote_test.go
jjbustamante: I addressed this suggestion in the following [commit](https://github.com/buildpacks/imgutil/pull/123/commits/d588cd706db0e32f55858fa00211c948d045210a)

remote/remote_test.go
natalieparellano: How about just `os.Unsetenv`?

jjbustamante: Oh! that one seems to be even better!!! Thanks!!!

remote/remote_test.go
jjbustamante: Done! added in the following [commit](https://github.com/buildpacks/imgutil/pull/123/commits/03a54599773fab9cb4690816c4dbe72252711908)

remote/remote_test.go
jjbustamante: > How about just `os.Unsetenv`?

I replaced the `it.Before` with the `os.Unsetenv` in the following [commit](https://github.com/buildpacks/imgutil/pull/123/commits/03a54599773fab9cb4690816c4dbe72252711908)

remote/remote_test.go
```diff
@@ -1500,4 +1500,100 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			})
+		})
+
+		when("image does not exist in the registry and client has read/write access", func() {
+			it("returns true", func() {
+				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+				h.AssertEq(t, image.CheckReadWriteAccess(), true)
+			})
+		})
```
natalieparellano, line 1581: This seems to be the most valuable "does not exist in the registry" case. I think the other "does not exist in the registry" cases might make more sense as "exists in the registry" cases. But maybe that's just me...

remote/remote_test.go
```diff
@@ -1500,4 +1500,100 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			})
+
+			it("returns true", func() {
+				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+				h.AssertEq(t, image.CheckReadWriteAccess(), true)
+			})
+		})
+
+		when("image does not exist in the registry and client only has read access", func() {
```
natalieparellano, line 1559: See https://github.com/buildpacks/imgutil/pull/123/files#r648687998

remote/remote_test.go
```diff
@@ -1500,4 +1500,100 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+
+		when("image does not exist in the registry and client has read/write access", func() {
+			it("returns true", func() {
+				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
+				h.AssertNil(t, err)
+				h.AssertEq(t, image.CheckReadWriteAccess(), true)
+			})
+		})
+
+		when("image does not exist in the registry and client doesn't have read/write access", func() {
```
natalieparellano, line 1583: See https://github.com/buildpacks/imgutil/pull/123/files#r648687998

natalieparellano (approved):
This looks great! I left a comment about inverting a couple "does not exist in the registry" cases (for readability), but it's just a suggestion.

PR #125, opened 2021-07-13
Adding a custom http handler for read/write access validation

jjbustamante:

## Summary

This PR is adding a custom HTTP handler to the docker registry to validate read/write access to some image name, these methods will be used to fix the issue [572 - validate registry credentials](https://github.com/buildpacks/lifecycle/issues/572) in the [lifecycle](https://github.com/buildpacks/lifecycle) repository

## Documentation

    - [x] No

## Related
Resolves #572 (in lifecycle)

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -356,7 +356,7 @@ func (i *Image) CheckReadWriteAccess() bool {
 	if err != nil {
 		return false
 	}
-	return remote.CheckPushPermission(ref, i.keychain, http.DefaultTransport) == nil
+	return i.CheckReadAccess() && remote.CheckPushPermission(ref, i.keychain, http.DefaultTransport) == nil
 }
 
 func (i *Image) CheckReadAccess() bool {
--- a/remote/remote_test.go
+++ b/remote/remote_test.go
@@ -23,10 +23,10 @@ import (
 var dockerRegistry, readonlyDockerRegistry, customRegistry *h.DockerRegistry
 
 const (
-	readWriteImage      = "image-readable-writable"
-	readNoWriteImage    = "image-readable"
-	noReadWriteImage    = "image-writable"
-	noReadNorWriteImage = "image-no-readable-no-writable"
+	readWriteImage = "image-readable-writable"
+	onlyReadImage  = "image-readable"
+	onlyWriteImage = "image-writable"
+	noAccessImage  = "noAccessImage"
 )
 
 func newTestImageName(providedPrefix ...string) string {
@@ -60,9 +60,9 @@ func TestRemote(t *testing.T) {
 
 	var customPrivileges = make(map[string]h.ImagePrivileges)
 	customPrivileges[readWriteImage] = h.NewImagePrivileges(readWriteImage)
-	customPrivileges[readNoWriteImage] = h.NewImagePrivileges(readNoWriteImage)
-	customPrivileges[noReadWriteImage] = h.NewImagePrivileges(noReadWriteImage)
-	customPrivileges[noReadNorWriteImage] = h.NewImagePrivileges("")
+	customPrivileges[onlyReadImage] = h.NewImagePrivileges(onlyReadImage)
+	customPrivileges[onlyWriteImage] = h.NewImagePrivileges(onlyWriteImage)
+	customPrivileges[noAccessImage] = h.NewImagePrivileges(noAccessImage)
 	customRegistry = h.NewDockerRegistry(h.WithAuth(customDockerConfigDir), h.WithSharedHandler(sharedRegistryHandler),
 		h.WithCustomPrivileges(customPrivileges))
 
@@ -1570,7 +1570,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				os.Setenv("DOCKER_CONFIG", dockerRegistry.DockerDirectory)
 			})
 
-			when("image has read/write access", func() {
+			when("image has read and write access", func() {
 				it("returns true", func() {
 					image, err := remote.NewImage(customRegistry.RepoName(readWriteImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
@@ -1580,7 +1580,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			when("image has read access but no write access", func() {
 				it("returns true", func() {
-					image, err := remote.NewImage(customRegistry.RepoName(readNoWriteImage), authn.DefaultKeychain)
+					image, err := remote.NewImage(customRegistry.RepoName(onlyReadImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
 					h.AssertEq(t, image.CheckReadAccess(), true)
 				})
@@ -1588,7 +1588,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			when("image doesn't have read access but has write access", func() {
 				it("returns false", func() {
-					image, err := remote.NewImage(customRegistry.RepoName(noReadWriteImage), authn.DefaultKeychain)
+					image, err := remote.NewImage(customRegistry.RepoName(onlyWriteImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
 					h.AssertEq(t, image.CheckReadAccess(), false)
 				})
@@ -1596,7 +1596,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			when("image doesn't have read nor write access", func() {
 				it("returns false", func() {
-					image, err := remote.NewImage(customRegistry.RepoName(noReadNorWriteImage), authn.DefaultKeychain)
+					image, err := remote.NewImage(customRegistry.RepoName(noAccessImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
 					h.AssertEq(t, image.CheckReadAccess(), false)
 				})
@@ -1668,7 +1668,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				os.Setenv("DOCKER_CONFIG", dockerRegistry.DockerDirectory)
 			})
 
-			when("image has read/write access", func() {
+			when("image has read and write access", func() {
 				it("returns true", func() {
 					image, err := remote.NewImage(customRegistry.RepoName(readWriteImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
@@ -1678,7 +1678,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			when("image has read access but no write access", func() {
 				it("returns false", func() {
-					image, err := remote.NewImage(customRegistry.RepoName(readNoWriteImage), authn.DefaultKeychain)
+					image, err := remote.NewImage(customRegistry.RepoName(onlyReadImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
 					h.AssertEq(t, image.CheckReadWriteAccess(), false)
 				})
@@ -1686,7 +1686,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			when("image doesn't have read access but has write access", func() {
 				it("returns true", func() {
-					image, err := remote.NewImage(customRegistry.RepoName(noReadWriteImage), authn.DefaultKeychain)
+					image, err := remote.NewImage(customRegistry.RepoName(onlyWriteImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
 					h.AssertEq(t, image.CheckReadWriteAccess(), false)
 				})
@@ -1694,7 +1694,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			when("image doesn't have read nor write access", func() {
 				it("returns false", func() {
-					image, err := remote.NewImage(customRegistry.RepoName(noReadNorWriteImage), authn.DefaultKeychain)
+					image, err := remote.NewImage(customRegistry.RepoName(noAccessImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
 					h.AssertEq(t, image.CheckReadWriteAccess(), false)
 				})
--- a/remote/remote_test.go
+++ b/remote/remote_test.go
@@ -20,7 +20,14 @@ import (
 	h "github.com/buildpacks/imgutil/testhelpers"
 )
 
-var dockerRegistry, readonlyDockerRegistry *h.DockerRegistry
+var dockerRegistry, readonlyDockerRegistry, customRegistry *h.DockerRegistry
+
+const (
+	readWriteImage      = "image-readable-writable"
+	readNoWriteImage    = "image-readable"
+	noReadWriteImage    = "image-writable"
+	noReadNorWriteImage = "image-no-readable-no-writable"
+)
 
 func newTestImageName(providedPrefix ...string) string {
 	prefix := "pack-image-test"
@@ -46,6 +53,20 @@ func TestRemote(t *testing.T) {
 	readonlyDockerRegistry = h.NewDockerRegistry(h.WithSharedHandler(sharedRegistryHandler))
 	readonlyDockerRegistry.Start(t)
 	defer readonlyDockerRegistry.Stop(t)
+
+	customDockerConfigDir, err := ioutil.TempDir("", "test.docker.config.custom.dir")
+	h.AssertNil(t, err)
+	defer os.RemoveAll(customDockerConfigDir)
+
+	var customPrivileges = make(map[string]h.ImagePrivileges)
+	customPrivileges[readWriteImage] = h.NewImagePrivileges(readWriteImage)
+	customPrivileges[readNoWriteImage] = h.NewImagePrivileges(readNoWriteImage)
+	customPrivileges[noReadWriteImage] = h.NewImagePrivileges(noReadWriteImage)
+	customPrivileges[noReadNorWriteImage] = h.NewImagePrivileges("")
+	customRegistry = h.NewDockerRegistry(h.WithAuth(customDockerConfigDir), h.WithSharedHandler(sharedRegistryHandler),
+		h.WithCustomPrivileges(customPrivileges))
+
+	customRegistry.Start(t)
 
 	os.Setenv("DOCKER_CONFIG", dockerRegistry.DockerDirectory)
 	defer os.Unsetenv("DOCKER_CONFIG")
@@ -1539,6 +1560,48 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				h.AssertEq(t, image.CheckReadAccess(), false)
 			})
 		})
+
+		when("using custom handler", func() {
+			it.Before(func() {
+				os.Setenv("DOCKER_CONFIG", customRegistry.DockerDirectory)
+			})
+
+			it.After(func() {
+				os.Setenv("DOCKER_CONFIG", dockerRegistry.DockerDirectory)
+			})
+
+			when("image has read/write access", func() {
+				it("returns true", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(readWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadAccess(), true)
+				})
+			})
+
+			when("image has read access but no write access", func() {
+				it("returns true", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(readNoWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadAccess(), true)
+				})
+			})
+
+			when("image doesn't have read access but has write access", func() {
+				it("returns false", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(noReadWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadAccess(), false)
+				})
+			})
+
+			when("image doesn't have read nor write access", func() {
+				it("returns false", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(noReadNorWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadAccess(), false)
+				})
+			})
+		})
 	})
 
 	when("#CheckReadWriteAccess", func() {
@@ -1593,6 +1656,48 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
 				h.AssertNil(t, err)
 				h.AssertEq(t, image.CheckReadWriteAccess(), false)
+			})
+		})
+
+		when("using custom handler", func() {
+			it.Before(func() {
+				os.Setenv("DOCKER_CONFIG", customRegistry.DockerDirectory)
+			})
+
+			it.After(func() {
+				os.Setenv("DOCKER_CONFIG", dockerRegistry.DockerDirectory)
+			})
+
+			when("image has read/write access", func() {
+				it("returns true", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(readWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadWriteAccess(), true)
+				})
+			})
+
+			when("image has read access but no write access", func() {
+				it("returns false", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(readNoWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadWriteAccess(), false)
+				})
+			})
+
+			when("image doesn't have read access but has write access", func() {
+				it("returns true", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(noReadWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadWriteAccess(), false)
+				})
+			})
+
+			when("image doesn't have read nor write access", func() {
+				it("returns false", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(noReadNorWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadWriteAccess(), false)
+				})
 			})
 		})
 	})
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -108,7 +108,7 @@ func Custom(handler http.Handler, permissions map[string]ImagePrivileges) http.H
 				}
 			} else {
 				if !permission.writable {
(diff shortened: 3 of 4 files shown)


testhelpers/docker_registry.go
```diff
@@ -11,25 +11,33 @@ import (
...
+	Host             string
+	Port             string
+	Name             string
+	server           *httptest.Server
+	DockerDirectory  string
+	username         string
+	password         string
+	regHandler       http.Handler
+	authnHandler     http.Handler
+	customHandler    http.Handler
```
natalieparellano, line 30: Do we need this? Is it set anywhere?

testhelpers/docker_registry.go
```diff
@@ -103,6 +148,9 @@ func (r *DockerRegistry) Start(t *testing.T) {
 	// wrap registry handler with authentication handler, defaulting to read-only
 	r.authnHandler = ReadOnly(r.regHandler)
 	if r.username != "" {
+		if r.customHandler != nil && r.customPrivileges != nil {
+			r.regHandler = Custom(r.regHandler, r.customPrivileges)
```
natalieparellano, line 152: Do you mean to be setting `r.authnHandler` here?

natalieparellano: (and if so, should we break after this line)

testhelpers/docker_registry.go
```diff
@@ -88,6 +103,36 @@ func ReadOnly(handler http.Handler) http.Handler {
...
 }
 
+func Custom(handler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if strings.Contains(r.URL.Path, "blobs") ||
+			strings.Contains(r.URL.Path, "manifests") ||
+			strings.Contains(r.URL.Path, "tags") {
+			elems := strings.Split(r.URL.Path, "/")
+			elems = elems[1:]
+			name := elems[len(elems)-1]
```
natalieparellano, line 113: A helper function might be useful here (to make the code more readable)

natalieparellano:
I love it! Left a couple nits & questions.

testhelpers/docker_registry.go
jjbustamante: No,  actually I am wrapping the regular handler with my custom behavior like this:
wrapping
The idea is to keep the basic authentication logic for the registry, then in the cases, I want to simulate the read/write behavior access I will do it in the custom and all other requests I do not care about will go to the normal registry logic. So, lines 151 and 152 are ok exactly like that.

testhelpers/docker_registry.go
jjbustamante: Oh. I just noticed I am not using it! let me removes it

testhelpers/docker_registry.go
aemengo: I'm having trouble here. `name` is supposed to be the image tag, correct? The same value passed into the `permissions` mapping?

The following is what I'm seeing for `r.URL.Path` with `remote.NewImage()`. Am I using it right?

```shell
# image tag: 192.168.1.164:63737/aemengo/test
Path: /v2/aemengo/test/blobs/sha256:60df352a35c779a36da39e670ddc90742f64209045365ea0ac4f5eb83422a615
Path: /v2/aemengo/test/blobs/uploads/
Path: /v2/aemengo/test/blobs/uploads/4831389563158288344
Path: /v2/aemengo/test/blobs/uploads/4831389563158288344
Path: /v2/aemengo/test/manifests/latest
Path: /v2/aemengo/test/manifests/latest
Path: /v2/aemengo/test/blobs/uploads/
Path: /v2/aemengo/test/blobs/uploads/261049867304784443
```

jjbustamante: well, in this case, `name` should be `test` and that should be the key to find in `permissions`. that you put before it is not going to be used to determine the permission (at least from the test perspective)

testhelpers/docker_registry.go
natalieparellano: Ohhh thank you for this explanation! That makes sense. Maybe add a comment?

natalieparellano:
```suggestion
			// wrap registry handler with custom handler
			r.regHandler = Custom(r.regHandler, r.customPrivileges)
```

testhelpers/docker_registry.go
natalieparellano: Hmm... this playground is giving me a blank name: https://play.golang.org/p/Zn2EqfiHGb5

I think we can't assume that `test` will be at the end of the path...

jjbustamante: Yeap, it's my mistake I will fix it ;) thanks @natalieparellano and @aemengo

jjbustamante: @aemengo , @natalieparellano 

What about something like [this](https://play.golang.org/p/QfbivVx2xIt) to parse the URL and extract the `name`, @aemengo I am trying to handle you mentioned 😄 .

I like the regex to isolate the string to V2/[whatever]/(blobs|manifest|tag), but I open to suggestion to get [whatever] in a optimal way

aemengo: @jjbustamante Why so complicated? 😅 Take a look at [this one](https://play.golang.org/p/F3b-dSljAQT)

jjbustamante: hahaha.. come on! I hate regex, thanks! any thoughts on best option to remove the /V2/****/(blobs|manifest|tag) stuff? and just keep the name?

aemengo: If you use the `FindStringSubmatch` method, the name will be 2nd value in an array of 2 elements. 👍🏾 

```
matches := s.FindStringSubmatch(str)
matches[1] //<--- name
```

jjbustamante: Awesome! I like it! thanks

testhelpers/docker_registry.go
jjbustamante: Added at the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/4261e2896afc176ea6a4967a4bd255f780340892)

testhelpers/docker_registry.go
jjbustamante: Removed in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/4261e2896afc176ea6a4967a4bd255f780340892)

testhelpers/docker_registry.go
jjbustamante: I fixed the issue and took into consideration the suggestion, the logic was moved to the helper function and I used the regex pattern matching to find the `name` see [commit](https://github.com/buildpacks/imgutil/pull/125/commits/4261e2896afc176ea6a4967a4bd255f780340892)

testhelpers/image_privileges.go
```diff
@@ -0,0 +1,35 @@
...
+func NewImagePrivileges(imageName string) *ImagePrivileges {
+	var isReadable, isWriteable bool
+	if strings.Contains(imageName, "readable") {
+		isReadable = true
+	}
+	if strings.Contains(imageName, "writable") {
+		isWriteable = true
+	}
+	return &ImagePrivileges{readable: isReadable, writeable: isWriteable}
+}
```
aemengo, line 24:
```suggestion
func NewImagePrivileges(imageName string) (priv ImagePrivileges) {
	priv.readable = strings.Contains(imageName, "readable")
	priv.writeable = strings.Contains(imageName, "writable")
	return
}
```

I liked the old interface better. But you're using it, not me 😅

aemengo (approved):

testhelpers/image_privileges.go
yaelharel: I don't see who is using this function but I think the `if` conditions here are redundant.
I'm ok with returning a pointer, but I would remove the conditions and just use the return value from `strings.Contains` to set the variables.

jjbustamante: I added this way of setting the properties based on @natalieparellano comments [here](https://github.com/buildpacks/lifecycle/pull/646#discussion_r669816609), I will use this method in the lifecycle unit tests

jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/b6d4eb2c54b191a59bf4a46c43c492a881e6889b)

testhelpers/docker_registry.go
```diff
@@ -17,15 +17,16 @@ import (
...
+	Host             string
+	Port             string
+	Name             string
+	server           *httptest.Server
+	DockerDirectory  string
+	username         string
+	password         string
+	regHandler       http.Handler
+	authnHandler     http.Handler
+	customPrivileges map[string]*ImagePrivileges
```
yaelharel: This is a map from an `imageName` to its permissions, right?
I think it's worth adding it as a comment

testhelpers/image_privileges.go
```diff
@@ -0,0 +1,35 @@
...
+
+func extractImageName(path string) string {
```
yaelharel, line 21: Can you please add a description to this function?

testhelpers/docker_registry.go
jjbustamante: Added in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/714d0d197edf3ff7391210592d4c3552cdfca10b)

testhelpers/image_privileges.go
jjbustamante: Added in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/714d0d197edf3ff7391210592d4c3552cdfca10b)

testhelpers/docker_registry.go
```diff
@@ -88,6 +96,27 @@ func ReadOnly(handler http.Handler) http.Handler {
 	})
 }
 
+func Custom(handler http.Handler, permissions map[string]*ImagePrivileges) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if permission, ok := permissions[extractImageName(r.URL.Path)]; ok {
+			if r.Method == "GET" || r.Method == "HEAD" {
+				if !permission.readable {
+					w.WriteHeader(401)
+					_, _ = w.Write([]byte("Unauthorized.\n"))
```
yaelharel, line 105: Looking at the return value of `ReadOnly` (although the use case is a bit different), I'm wondering whether it should be 401 or 405 ("Method Not Allowed")? I'm not sure which one is more accurate in this case.

testhelpers/docker_registry.go
```diff
@@ -88,6 +96,27 @@ func ReadOnly(handler http.Handler) http.Handler {
 	})
 }
 
+func Custom(handler http.Handler, permissions map[string]*ImagePrivileges) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if permission, ok := permissions[extractImageName(r.URL.Path)]; ok {
+			if r.Method == "GET" || r.Method == "HEAD" {
+				if !permission.readable {
```
yaelharel, line 103: I would add a comment in each of these conditions to make the flow clearer. Something like:
```
// read request was arrived while read permission isn't allowed
```

yaelharel (commented):
I think it's worth adding some tests of this new functionality.

testhelpers/docker_registry.go
```diff
@@ -17,15 +17,16 @@ import (
...
+	Host             string
+	Port             string
+	Name             string
+	server           *httptest.Server
+	DockerDirectory  string
+	username         string
+	password         string
+	regHandler       http.Handler
+	authnHandler     http.Handler
+	customPrivileges map[string]ImagePrivileges // map from an imageNames to its permissions
```
yaelharel, line 29:
```suggestion
	customPrivileges map[string]ImagePrivileges // map from an imageName to its permissions
```

testhelpers/docker_registry.go
jjbustamante: Well, that's actually one of the problems with the ReadOnly when we try to validate read/write access, we are expecting to receive `401 Unauthorized` if your credentials are wrong or you don't have access to resources.

testhelpers/docker_registry.go
jjbustamante: Done. [commit](https://github.com/buildpacks/imgutil/pull/125/commits/a0fd89a42b3519eaf6dc613931da5b34d11cf397)

testhelpers/docker_registry.go
jjbustamante: Done [commit](https://github.com/buildpacks/imgutil/pull/125/commits/a0fd89a42b3519eaf6dc613931da5b34d11cf397)

testhelpers/docker_registry.go
yaelharel: So why don't we fix the `ReadOnly` function?

testhelpers/docker_registry.go
```diff
@@ -88,6 +96,29 @@ func ReadOnly(handler http.Handler) http.Handler {
...
+			if r.Method == "GET" || r.Method == "HEAD" {
+				if !permission.readable {
+					// read request was arrived while read permission isn't allowed
+					w.WriteHeader(401)
+					_, _ = w.Write([]byte("Unauthorized.\n"))
+					return
+				}
+			} else {
+				if !permission.writable {
+					// write request was arrived while read permission isn't allowed
```
yaelharel, line 111:
```suggestion
					// write request was arrived while write permission isn't allowed
```

jjbustamante: Fixed [commit](https://github.com/buildpacks/imgutil/pull/125/commits/6a95fbcf1491910ebe83969ffeec0d73807786de) 

Sorry!

jjbustamante:
> I think it's worth adding some tests of this new functionality.

I just added more unit tests to the read/write methods using the custom handler. I noticed that I think the `checkReadWriteAccess` was not validating read access, I will like to take a look at that test case (not sure if that is even possible, having write access but not read, I think @natalieparellano mentioned that before). Any thoughts will be awesome.

testhelpers/docker_registry.go
jjbustamante: There are several tests that depend on that, it could be an option, but actually not sure how good or bad it could be, but in any case, this docker registry implementation is a custom implementation created just to handle test scenarios. From my perspective having a version of a registry that behaves in a certain way for given scenarios is not that bad

natalieparellano: IIUC we expect Docker Hub to return a 401, therefore the test registry should return a 401.

remote/remote.go
```diff
@@ -356,7 +356,7 @@ func (i *Image) CheckReadWriteAccess() bool {
 	if err != nil {
 		return false
 	}
-	return remote.CheckPushPermission(ref, i.keychain, http.DefaultTransport) == nil
+	return i.CheckReadAccess() && remote.CheckPushPermission(ref, i.keychain, http.DefaultTransport) == nil
```
natalieparellano, line 359: I agree that "write only" access would be silly. But this doesn't do any harm...

testhelpers/image_privileges.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+var urlRegex = regexp.MustCompile(`v2\/(.*)\/(?:blobs|manifests|tags)`)
+
+type ImagePrivileges struct {
+	readable bool
+	writable bool
+}
+
+func NewImagePrivileges(imageName string) (priv ImagePrivileges) {
```
natalieparellano, line 15: Could we add a docs comment here? Keying the permissions off the image name (while convenient) wasn't immediately obvious to me.

remote/remote_test.go
```diff
@@ -1595,5 +1658,47 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadWriteAccess(), true)
+				})
+			})
+
+			when("image has read access but no write access", func() {
+				it("returns false", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(readNoWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadWriteAccess(), false)
```
natalieparellano, line 1683: Should we check read access here?

remote/remote_test.go
```diff
@@ -1595,5 +1658,47 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+				})
+			})
+
+			when("image doesn't have read nor write access", func() {
+				it("returns false", func() {
+					image, err := remote.NewImage(customRegistry.RepoName(noReadNorWriteImage), authn.DefaultKeychain)
+					h.AssertNil(t, err)
+					h.AssertEq(t, image.CheckReadWriteAccess(), false)
```
natalieparellano, line 1699: Should we check read access here?

natalieparellano:
@jjbustamante thank you for adding the tests! This looks great!

natalieparellano (approved):

remote/remote_test.go
jjbustamante: Do you mean adding the assert `image.CheckReadAccess()`? because that one seems to be tested at line 1585

remote/remote_test.go
jjbustamante: Thinking a lit bit more about it, with these new docker registry handlers, I can actually cut in half the code and added the assertions for `CheckReadWriteAccess()` and `CheckReadAccess()` in the same test scenarios. Would it help on making the test cases more readable?

remote/remote.go
jjbustamante: Actually, because we changed the method signature to `CheckReadWriteAccess()`, actually it should also validate the Read access, right?

remote/remote_test.go
natalieparellano: Ohh I missed that these were all scoped to `CheckReadAccess()`. I think it's okay.

remote/remote_test.go
natalieparellano: Oh, maybe! Up to you...

testhelpers/docker_registry.go
yaelharel: @natalieparellano, it makes sense that it'll return 401. My question is why `ReadOnly` returns 405 🤔

testhelpers/image_privileges.go
jjbustamante: Added in this [commit](https://github.com/buildpacks/imgutil/pull/125/commits/2486b56700226982aa19e6e383bba488b80e0394)

testhelpers/image_privileges.go
```diff
@@ -0,0 +1,41 @@
...
+// For examples:
+// NewImagePrivileges("") returns ImagePrivileges{readable: false, writable: false}
+// NewImagePrivileges("foo-readable") returns ImagePrivileges{readable: true, writable: false}
+// NewImagePrivileges("foo-writable") returns ImagePrivileges{readable: false, writable: true}
+// NewImagePrivileges("foo-writable-readable") returns ImagePrivileges{readable: true, writable: true}
+func NewImagePrivileges(imageName string) (priv ImagePrivileges) {
+	priv.readable = strings.Contains(imageName, "readable")
+	priv.writable = strings.Contains(imageName, "writable")
+	return
+}
```
yaelharel, line 25: Maybe I'm wrong but according to what I saw in our code, we usually don't set the returned variable name in the function definition. We define it as part of the function and return it explicitly.
If I'm right, maybe it's better to change the function to:

```suggestion
func NewImagePrivileges(imageName string) (priv ImagePrivileges) {
	image ImagePrivileges
	image.readable = strings.Contains(imageName, "readable")
	image.writable = strings.Contains(imageName, "writable")
	return image
}
```

remote/remote_test.go
```diff
@@ -20,7 +20,14 @@ import (
 	h "github.com/buildpacks/imgutil/testhelpers"
 )
 
-var dockerRegistry, readonlyDockerRegistry *h.DockerRegistry
+var dockerRegistry, readonlyDockerRegistry, customRegistry *h.DockerRegistry
+
+const (
+	readWriteImage      = "image-readable-writable"
+	readNoWriteImage    = "image-readable"
```
yaelharel, line 27:
```suggestion
	onlyReadImage    = "image-readable"
```

remote/remote_test.go
```diff
@@ -20,7 +20,14 @@ import (
...
+	noReadWriteImage    = "image-writable"
```
yaelharel, line 28:
```suggestion
	onlyWriteImage    = "image-writable"
```

remote/remote_test.go
```diff
@@ -20,7 +20,14 @@ import (
...
+	noReadNorWriteImage = "image-no-readable-no-writable"
```
yaelharel, line 29:
```suggestion
	noAccessImage = "no-access-image"
```

remote/remote_test.go
```diff
@@ -47,6 +54,20 @@ func TestRemote(t *testing.T) {
 	readonlyDockerRegistry.Start(t)
 	defer readonlyDockerRegistry.Stop(t)
 
+	customDockerConfigDir, err := ioutil.TempDir("", "test.docker.config.custom.dir")
+	h.AssertNil(t, err)
+	defer os.RemoveAll(customDockerConfigDir)
+
+	var customPrivileges = make(map[string]h.ImagePrivileges)
```
yaelharel, line 61: Maybe you can define all of the consts in a map, and then loop on that map and initialize the `customPrivileges`.
In this way, if we'll decide to add/remove images, we'll need to do it only in one place.

remote/remote_test.go
```diff
@@ -1539,6 +1560,48 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+		when("using custom handler", func() {
+			it.Before(func() {
+				os.Setenv("DOCKER_CONFIG", customRegistry.DockerDirectory)
+			})
+
+			it.After(func() {
+				os.Setenv("DOCKER_CONFIG", dockerRegistry.DockerDirectory)
+			})
+
+			when("image has read/write access", func() {
```
yaelharel, line 1573:
```suggestion
			when("image has read and write access", func() {
```

yaelharel: Why do we only call `CheckReadAccess` instead of `CheckReadWriteAccess`?
Moreover, maybe it's better that `CheckReadWriteAccess` will return 2 boolean values for the read and write access separately.

remote/remote_test.go
```diff
@@ -1595,5 +1658,47 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
```
yaelharel, line 1671:
```suggestion
			when("image has read and write access", func() {
```

remote/remote_test.go
jjbustamante: Added in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/87d5ba9fc355d30923e2cf672f08e02c9de908fd)

remote/remote_test.go
jjbustamante: Added in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/87d5ba9fc355d30923e2cf672f08e02c9de908fd)

remote/remote_test.go
jjbustamante: Added in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/87d5ba9fc355d30923e2cf672f08e02c9de908fd)

remote/remote_test.go
jjbustamante: Added in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/87d5ba9fc355d30923e2cf672f08e02c9de908fd)

remote/remote_test.go
jjbustamante: Added in the following [commit](https://github.com/buildpacks/imgutil/pull/125/commits/87d5ba9fc355d30923e2cf672f08e02c9de908fd)

remote/remote.go
yaelharel: I have some thoughts about the `CheckReadWriteAccess` and `CheckReadAccess` functions:
* Maybe we should move `remote.CheckPushPermission(ref, i.keychain, http.DefaultTransport) == nil` to a private function `checkWriteAccess` so the code will be clearer.
* Should `CheckReadWriteAccess` return two booleans or a struct with two booleans so it'll be clearer what kind of access (if any) the image has?
* Although both functions are public, I think it's redundant to run almost the same tests as part of `remote_test.go`. Should we only keep those under `CheckReadWriteAccess`? And if not, ,maybe we can create a function to test all of these scenarios together (the function will get the function that should run on the image - either `CHeckReadWriteAccess` or `CheckReadAccess`).

jjbustamante: Regarding your third comment, I thought something similar [here](https://github.com/buildpacks/imgutil/pull/125#discussion_r671434624). This is a preview of how it will look if I move the `CheckReadAccess` assertions to the `CheckReadWriteAccess` scenarios

```
when("image has read and write access", func() {
    it("returns true", func() {
	    image, err := remote.NewImage(customRegistry.RepoName(readWriteImage), authn.DefaultKeychain)
	    h.AssertNil(t, err)
	    h.AssertEq(t, image.CheckReadWriteAccess(), true)
	    h.AssertEq(t, image.CheckReadAccess(), true)
    })
})

when("image has read access but no write access", func() {
    it("returns true for CheckReadAccess() and false for CheckReadWriteAccess()", func() {
	    image, err := remote.NewImage(customRegistry.RepoName(onlyReadImage), authn.DefaultKeychain)
	    h.AssertNil(t, err)
	    h.AssertEq(t, image.CheckReadWriteAccess(), false)
	    h.AssertEq(t, image.CheckReadAccess(), true)
    })
})

when("image doesn't have read access but has write access", func() {
    it("returns false", func() {
	    image, err := remote.NewImage(customRegistry.RepoName(onlyWriteImage), authn.DefaultKeychain)
	    h.AssertNil(t, err)
	    h.AssertEq(t, image.CheckReadWriteAccess(), false)
	    h.AssertEq(t, image.CheckReadAccess(), false)
    })
})

when("image doesn't have read nor write access", func() {
    it("returns false", func() {
	    image, err := remote.NewImage(customRegistry.RepoName(noAccessImage), authn.DefaultKeychain)
	    h.AssertNil(t, err)
	    h.AssertEq(t, image.CheckReadWriteAccess(), false)
	    h.AssertEq(t, image.CheckReadAccess(), false)
    })
})
```

yaelharel: I'm wondering if it doesn't make the code confusing, because all of these `when`s will be under `when("CheckReadWriteAccess"`, right?
I thought about completely deleting the `CheckReadAccess` calls because they're part of the `CheckReadWriteAccess` code, but then I think that we need to change a bit the `CheckReadWriteAccess` interface to return what kind of access we have.

testhelpers/docker_registry.go
natalieparellano: I am not sure - IIRC @micahyoung had done some investigation on what the failure response looks like for different registries. The [distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes) is a bit ambiguous on what specific code should be returned. I would not be opposed to updating `ReadOnly`, but I don't think it needs to block this PR.

yaelharel: Of course.
Maybe it's worth opening a separate issue for it - fix the errors that should be returned in each case (so at least we'll be consistent).

PR #126, opened 2021-07-28
Test helpers improvements

jjbustamante:
### References

- This Pull Request is created after [PR #125](https://github.com/buildpacks/imgutil/pull/125) was merged

### Context

The idea behind [PR #125](https://github.com/buildpacks/imgutil/pull/125) was to provide a structure to help us on testing the new validations in the lifecycle described in issue [#572](https://github.com/buildpacks/lifecycle/issues/572). At that point, some improvements were raised and the idea was to simply our docker registry implementation. This pull request is addressing those improvements.

### Description

- The PR exposes a new method `Add` to allow dynamically configure the `ImagePrivileges` in the registry, basically providing an option to include new mappings after the registry was created.
- The PR is also addressing one little problem that avoid us from using only one docker registry implementation to do our tests.

### Actual

Right now, if we need to simulate a read-only registry or a registry based on basic authentication, we need to create two instances similar  to the following image

test-helpers-improvements-1

Basically, if the registry is read-only no authentication check is made. After merging the [PR #125](https://github.com/buildpacks/imgutil/pull/125), we added a third behavior

wrapping
 
As we can see, the problem that avoid us from using the custom registry to implement both behaviors (read-only and basic authentication) is that for the case of read-only we are actually running credential validation!

### After

The PR is implementing the following behavior 

registry-3

We are improving the Custom logic to handle the scenarios:
 * The image is readable
 * The image is writable 
 * Or both
 In this way, we are going to be able to simplify the tests in the lifecycle by removing the docker registry instances and using only one to handle all the test cases

### DIFF
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -200,7 +200,7 @@ func (r *DockerRegistry) EncodedLabeledAuth() string {
 }
 
 // Add stores the given key name with the provided ImagePrivileges. For example
-// Add("my-image", NewImagePrivileges("foo-writable-readable")) will save "my-image" as a
+// Add("my-image", NewImagePrivileges(Readable, Writable)) will save "my-image" as a
 // readable and writable image into the registry.
 // It must be called before Start to have any effect.
 func (r *DockerRegistry) Add(key string, privilege ImagePrivileges) {
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -97,27 +97,28 @@ func ReadOnly(handler http.Handler) http.Handler {
 
 func delegator(basicAuthHandler http.Handler, regHandler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if permission, ok := permissions[extractImageName(r.URL.Path)]; ok {
-			if r.Method == "GET" || r.Method == "HEAD" {
-				if permission.readable {
-					regHandler.ServeHTTP(w, r)
-				} else {
-					// read request was arrived while read permission isn't allowed
-					w.WriteHeader(401)
-					_, _ = w.Write([]byte("Unauthorized.\n"))
-				}
-			} else {
-				if permission.writable {
-					regHandler.ServeHTTP(w, r)
-				} else {
-					// write request was arrived while write permission isn't allowed
-					w.WriteHeader(401)
-					_, _ = w.Write([]byte("Unauthorized.\n"))
-				}
-			}
+		var (
+			image ImagePrivileges
+			ok    bool
+		)
+		if image, ok = permissions[extractImageName(r.URL.Path)]; !ok {
+			basicAuthHandler.ServeHTTP(w, r)
 			return
 		}
-		basicAuthHandler.ServeHTTP(w, r)
+		if r.Method == "GET" || r.Method == "HEAD" {
+			if !image.readable {
+				w.WriteHeader(401)
+				_, _ = w.Write([]byte("Unauthorized.\n"))
+				return
+			}
+		} else { // assume write request
+			if !image.writable {
+				w.WriteHeader(401)
+				_, _ = w.Write([]byte("Unauthorized.\n"))
+				return
+			}
+		}
+		regHandler.ServeHTTP(w, r)
 	})
 }
 
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -84,14 +84,14 @@ func BasicAuth(handler http.Handler, username, password, realm string) http.Hand
 
 // ReadOnly wraps a handler, allowing only GET and HEAD requests, otherwise rejecting with a 405
 func ReadOnly(handler http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, request *http.Request) {
+	return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
 		if !isReadRequest(request) {
-			w.WriteHeader(405)
-			_, _ = w.Write([]byte("Method Not Allowed.\n"))
+			response.WriteHeader(405)
+			_, _ = response.Write([]byte("Method Not Allowed.\n"))
 			return
 		}
 
-		handler.ServeHTTP(w, request)
+		handler.ServeHTTP(response, request)
 	})
 }
 
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -84,41 +84,41 @@ func BasicAuth(handler http.Handler, username, password, realm string) http.Hand
 
 // ReadOnly wraps a handler, allowing only GET and HEAD requests, otherwise rejecting with a 405
 func ReadOnly(handler http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if !(r.Method == "GET" || r.Method == "HEAD") {
+	return http.HandlerFunc(func(w http.ResponseWriter, request *http.Request) {
+		if !isReadRequest(request) {
 			w.WriteHeader(405)
 			_, _ = w.Write([]byte("Method Not Allowed.\n"))
 			return
 		}
 
-		handler.ServeHTTP(w, r)
+		handler.ServeHTTP(w, request)
 	})
 }
 
 func delegator(basicAuthHandler http.Handler, regHandler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
 		var (
 			image ImagePrivileges
 			ok    bool
 		)
-		if image, ok = permissions[extractImageName(r.URL.Path)]; !ok {
-			basicAuthHandler.ServeHTTP(w, r)
+		if image, ok = permissions[extractImageName(request.URL.Path)]; !ok {
+			basicAuthHandler.ServeHTTP(response, request)
 			return
 		}
-		if r.Method == "GET" || r.Method == "HEAD" {
+		if isReadRequest(request) {
 			if !image.readable {
-				w.WriteHeader(401)
-				_, _ = w.Write([]byte("Unauthorized.\n"))
+				response.WriteHeader(401)
+				_, _ = response.Write([]byte("Unauthorized.\n"))
 				return
 			}
 		} else { // assume write request
 			if !image.writable {
-				w.WriteHeader(401)
-				_, _ = w.Write([]byte("Unauthorized.\n"))
+				response.WriteHeader(401)
+				_, _ = response.Write([]byte("Unauthorized.\n"))
 				return
 			}
 		}
-		regHandler.ServeHTTP(w, r)
+		regHandler.ServeHTTP(response, request)
 	})
 }
 
@@ -193,37 +193,49 @@ func (r *DockerRegistry) EncodedLabeledAuth() string {
 	return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(`{"username":"%s","password":"%s"}`, r.username, r.password)))
 }
 
-// SetPrivilegesToImage sets the given image name with the provided ImagePrivileges if ImagePrivileges was enabled.
+// setImagePrivileges sets the given image name with the provided ImagePrivileges if WithImagePrivileges was called during registry creation.
 // For example SetPrivilegesToImage("my-image", NewImagePrivileges(Readable, Writable)) will save "my-image" as a
 // readable and writable image into the registry.
-func (r *DockerRegistry) setPrivilegesToImage(imageName string, privilege ImagePrivileges) {
-	if r.isImagePrivilegesEnable() {
+func (r *DockerRegistry) setImagePrivileges(imageName string, privilege ImagePrivileges) {
+	if r.imagePrivilegesEnabled() {
 		r.imagePrivileges[imageName] = privilege
 	}
 }
 
-// MakeReadOnly set the given image name to be readable when the ImagePrivileges feature was enabled
-func (r *DockerRegistry) MakeReadOnly(imageName string) {
-	r.setPrivilegesToImage(imageName, NewImagePrivileges(Readable))
+// SetReadOnly set the given image name to be readable when the ImagePrivileges feature was enabled
+// Returns RepoName(imageName)
+func (r *DockerRegistry) SetReadOnly(imageName string) string {
+	r.setImagePrivileges(imageName, NewImagePrivileges(Readable))
+	return r.RepoName(imageName)
 }
 
-// MakeWriteOnly set the given image name to be writable when the ImagePrivileges feature was enabled
-func (r *DockerRegistry) MakeWriteOnly(imageName string) {
-	r.setPrivilegesToImage(imageName, NewImagePrivileges(Writable))
+// SetWriteOnly set the given image name to be writable when the ImagePrivileges feature was enabled
+// Returns RepoName(imageName)
+func (r *DockerRegistry) SetWriteOnly(imageName string) string {
+	r.setImagePrivileges(imageName, NewImagePrivileges(Writable))
+	return r.RepoName(imageName)
 }
 
-// MakeReadWrite set the given image name to be readable and writable when the ImagePrivileges feature was enabled
-func (r *DockerRegistry) MakeReadWrite(imageName string) {
-	r.setPrivilegesToImage(imageName, NewImagePrivileges(Readable, Writable))
+// SetReadWrite set the given image name to be readable and writable when the ImagePrivileges feature was enabled
+// Returns RepoName(imageName)
+func (r *DockerRegistry) SetReadWrite(imageName string) string {
+	r.setImagePrivileges(imageName, NewImagePrivileges(Readable, Writable))
+	return r.RepoName(imageName)
 }
 
-// MakeInaccessible set the given image name to do not have any access when the ImagePrivileges feature was enabled
-func (r *DockerRegistry) MakeInaccessible(imageName string) {
-	r.setPrivilegesToImage(imageName, NewImagePrivileges())
+// SetInaccessible set the given image name to do not have any access when the ImagePrivileges feature was enabled
+// Returns RepoName(imageName)
+func (r *DockerRegistry) SetInaccessible(imageName string) string {
+	r.setImagePrivileges(imageName, NewImagePrivileges())
+	return r.RepoName(imageName)
 }
 
-func (r *DockerRegistry) isImagePrivilegesEnable() bool {
+func (r *DockerRegistry) imagePrivilegesEnabled() bool {
 	return r.imagePrivileges != nil
+}
+
+func isReadRequest(req *http.Request) bool {
+	return req.Method == "GET" || req.Method == "HEAD"
 }
 
 //DockerHostname discovers the appropriate registry hostname.
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -107,7 +107,9 @@ func Custom(basicAuthHandler http.Handler, regHandler http.Handler, permissions 
 					return
 				}
 			} else {
-				if !permission.writable {
+				if permission.writable {
+					regHandler.ServeHTTP(w, r)
+				} else {
 					// write request was arrived while write permission isn't allowed
 					w.WriteHeader(401)
 					_, _ = w.Write([]byte("Unauthorized.\n"))
@@ -141,9 +143,9 @@ func (r *DockerRegistry) Start(t *testing.T) {
 	if r.username != "" {
 		if r.customPrivileges != nil {
 			// wrap registry handler with basic auth
-			r.authnHandler = BasicAuth(r.regHandler, r.username, r.password, "registry")
+			basicAuthHandler := BasicAuth(r.regHandler, r.username, r.password, "registry")
 			// wrap registry handler with custom handler
-			r.authnHandler = Custom(r.authnHandler, r.regHandler, r.customPrivileges)
+			r.authnHandler = Custom(basicAuthHandler, r.regHandler, r.customPrivileges)
 		} else {
 			// wrap registry handler basic auth
 			r.authnHandler = BasicAuth(r.regHandler, r.username, r.password, "registry")
--- a/testhelpers/docker_registry.go
+++ b/testhelpers/docker_registry.go
@@ -17,16 +17,16 @@ import (
 )
 
 type DockerRegistry struct {
-	Host             string
-	Port             string
-	Name             string
-	server           *httptest.Server
-	DockerDirectory  string
-	username         string
-	password         string
-	regHandler       http.Handler
-	authnHandler     http.Handler
-	customPrivileges map[string]ImagePrivileges // map from an imageName to its permissions
+	Host            string
+	Port            string
+	Name            string
+	server          *httptest.Server
+	DockerDirectory string
+	username        string
+	password        string
+	regHandler      http.Handler
+	authnHandler    http.Handler
+	imagePrivileges map[string]ImagePrivileges // map from an imageName to its permissions
 }
 
 type RegistryOption func(registry *DockerRegistry)
@@ -39,10 +39,11 @@ func WithSharedHandler(handler http.Handler) RegistryOption {
 	}
 }
 
-//WithCustomPrivileges allows to execute some customer read/write access validations based on the image name
-func WithCustomPrivileges(permissions map[string]ImagePrivileges) RegistryOption {
+//WithImagePrivileges enables the execution of read/write access validations based on the image name
+func WithImagePrivileges() RegistryOption {
+	var permissions = make(map[string]ImagePrivileges)
 	return func(registry *DockerRegistry) {
-		registry.customPrivileges = permissions
+		registry.imagePrivileges = permissions
 	}
 }
 
@@ -94,7 +95,7 @@ func ReadOnly(handler http.Handler) http.Handler {
 	})
 }
 
-func Custom(basicAuthHandler http.Handler, regHandler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
+func delegator(basicAuthHandler http.Handler, regHandler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if permission, ok := permissions[extractImageName(r.URL.Path)]; ok {
 			if r.Method == "GET" || r.Method == "HEAD" {
@@ -104,7 +105,6 @@ func Custom(basicAuthHandler http.Handler, regHandler http.Handler, permissions 
 					// read request was arrived while read permission isn't allowed
 					w.WriteHeader(401)
 					_, _ = w.Write([]byte("Unauthorized.\n"))
-					return
 				}
 			} else {
 				if permission.writable {
@@ -113,9 +113,9 @@ func Custom(basicAuthHandler http.Handler, regHandler http.Handler, permissions 
 					// write request was arrived while write permission isn't allowed
 					w.WriteHeader(401)
 					_, _ = w.Write([]byte("Unauthorized.\n"))
-					return
 				}
 			}
+			return
 		}
 		basicAuthHandler.ServeHTTP(w, r)
 	})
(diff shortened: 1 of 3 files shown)


testhelpers/docker_registry.go
```diff
@@ -68,17 +68,24 @@ func NewDockerRegistry(ops ...RegistryOption) *DockerRegistry {
 }
 
 // BasicAuth wraps a handler, allowing requests with matching username and password headers, otherwise rejecting with a 401
-func BasicAuth(handler http.Handler, username, password, realm string) http.Handler {
+// optPermissions is used to skip the authentication check if the image is readable.
```
natalieparellano, line 71:
```suggestion
// optPermissions is used to skip the authentication check if the image has been configured in the test setup to be always readable.
```

testhelpers/docker_registry.go
```diff
@@ -119,6 +126,11 @@ func Custom(handler http.Handler, permissions map[string]ImagePrivileges) http.H
 	})
 }
 
+// Start creates a docker registry following these rules:
+// - Shared handler will be use, otherwise a new one will be created
```
natalieparellano, line 130:
```suggestion
// - Shared handler will be used, otherwise a new one will be created
```

testhelpers/docker_registry.go
```diff
@@ -119,6 +126,11 @@ func Custom(handler http.Handler, permissions map[string]ImagePrivileges) http.H
...
+// - by default the shared handler will be wrapped using a read only handler
```
natalieparellano, line 131:
```suggestion
// - By default the shared handler will be wrapped using a read only handler
```

testhelpers/docker_registry.go
```diff
@@ -119,6 +126,11 @@ func Custom(handler http.Handler, permissions map[string]ImagePrivileges) http.H
...
+// - In case credentials are configured, the shared handler will be wrapped with a basic authentication handler and
+//   in any image privileges were set, then custom handler will be use to wrap the auth handler.
```
natalieparellano, line 133:
```suggestion
//   if any image privileges were set, then the custom handler will be used to wrap the auth handler.
```

testhelpers/docker_registry.go
```diff
@@ -68,17 +68,24 @@ func NewDockerRegistry(ops ...RegistryOption) *DockerRegistry {
 }
 
 // BasicAuth wraps a handler, allowing requests with matching username and password headers, otherwise rejecting with a 401
-func BasicAuth(handler http.Handler, username, password, realm string) http.Handler {
+// optPermissions is used to skip the authentication check if the image is readable.
+func BasicAuth(handler http.Handler, username, password, realm string, optPermissions ...map[string]ImagePrivileges) http.Handler {
```
natalieparellano, line 72: Why does `optPermissions` need to be passed in if `customPrivileges` is a field on the registry?

testhelpers/image_privileges.go
```diff
@@ -12,7 +12,7 @@ type ImagePrivileges struct {
 	writable bool
 }
 
-// Creates a new ImagePrivileges, use "readable" or "writable" to set the properties accordingly.
+// NewImagePrivileges Creates a new ImagePrivileges, use "readable" or "writable" to set the properties accordingly.
```
natalieparellano, line 15:
```suggestion
// NewImagePrivileges creates a new ImagePrivileges, use "readable" or "writable" to set the properties accordingly.
```

testhelpers/image_privileges.go
```diff
@@ -24,11 +24,11 @@ func NewImagePrivileges(imageName string) (priv ImagePrivileges) {
 	return
 }
 
-// Based of the registry API specification https://docs.docker.com/registry/spec/api/
-// This method returns the image name for path value that matches requests to blobs, manifests or tags
+// extractImageName returns the image name for path value that matches requests to blobs, manifests or tags
```
natalieparellano, line 27:
```suggestion
// extractImageName returns the image name from a path value that matches requests to blobs, manifests or tags
```

testhelpers/image_privileges.go
```diff
@@ -24,11 +24,11 @@ func NewImagePrivileges(imageName string) (priv ImagePrivileges) {
...
 // For examples:
 // extractImageName("v2/foo.bar/blobs/") returns "foo.bar"
 // extractImageName("v2/foo/bar/manifests/") returns "foo/bar"
+// Based of the registry API specification https://docs.docker.com/registry/spec/api/
```
natalieparellano, line 31:
```suggestion
// Based on the Docker registry API specification: https://docs.docker.com/registry/spec/api/
```

natalieparellano (commented):
@jjbustamante this will be a great improvement :)

Added a couple questions & nits.

testhelpers/docker_registry.go
jjbustamante: Because these HTTP handler functions don't have access to those attributes, they are plain functions handling the HTTP request. I think that's why we are also sending the username/password values from the registry.

natalieparellano: Ah so again, reading more closely, it would be called before `Start()`.

testhelpers/docker_registry.go
```diff
@@ -182,6 +198,13 @@ func (r *DockerRegistry) EncodedLabeledAuth() string {
 	return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(`{"username":"%s","password":"%s"}`, r.username, r.password)))
 }
 
+// Add stores the given key name with the provided ImagePrivileges. For example
+// Add("my-image", NewImagePrivileges("foo-writable-readable")) will save "my-image" as a
+// readable and writable image into the registry
```
natalieparellano, line 203:
```suggestion
// readable and writable image into the registry.
// It must be called before Start to have any effect.
```

testhelpers/docker_registry.go
jjbustamante: Actually, if I am not wrong, the HTTP server is holding the reference to the handler here:
```	
r.server = &httptest.Server{
	Listener: listener,
	Config:   &http.Server{Handler: r.authnHandler},
}
```
every time a new request is coming, the HTTP server is invoking the handler function which is, in fact, referencing the attribute from the docker registry (the permission map), and that is how it is dynamically updated when call the `Add` method  after the `Start`

testhelpers/image_privileges.go
natalieparellano: If I'm understanding https://github.com/buildpacks/lifecycle/pull/685/files correctly, we don't actually care that `imageName` is an image name. We just care what permissions we want the image to have when we call `testRegistry.Add(myImage, ih.NewImagePrivileges(readWriteImage)))`.

If that's true why not make `readable` and `writable` constants rather than relying on this fuzzy matching? (Here's a playground showing what I'm thinking: https://play.golang.org/p/tR0h1HFW8P- )

testhelpers/docker_registry.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1a4373ac85d225b6bd3771c3ccadae2a239cf6db)

testhelpers/docker_registry.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1a4373ac85d225b6bd3771c3ccadae2a239cf6db)

testhelpers/docker_registry.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1a4373ac85d225b6bd3771c3ccadae2a239cf6db)

testhelpers/docker_registry.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1a4373ac85d225b6bd3771c3ccadae2a239cf6db)

testhelpers/image_privileges.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1a4373ac85d225b6bd3771c3ccadae2a239cf6db)

testhelpers/image_privileges.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1a4373ac85d225b6bd3771c3ccadae2a239cf6db)

testhelpers/docker_registry.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1a4373ac85d225b6bd3771c3ccadae2a239cf6db)

testhelpers/image_privileges.go
jjbustamante: Yeah, I like it!

jjbustamante: I added those changes in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1924767685e509286c1db17a8e2c2f791113a114)

testhelpers/image_privileges.go
```diff
@@ -12,23 +12,36 @@ type ImagePrivileges struct {
...
-func NewImagePrivileges(imageName string) (priv ImagePrivileges) {
-	priv.readable = strings.Contains(imageName, "readable")
-	priv.writable = strings.Contains(imageName, "writable")
+// NewImagePrivileges() returns ImagePrivileges{readable: false, writable: false}
+// NewImagePrivileges(Readable) returns ImagePrivileges{readable: true, writable: false}
+// NewImagePrivileges(Writable) returns ImagePrivileges{readable: false, writable: true}
+// NewImagePrivileges(Readable, Writable) returns ImagePrivileges{readable: true, writable: true}
+func NewImagePrivileges(imageAccess ...ImageAccess) (priv ImagePrivileges) {
+	for _, ia := range imageAccess {
+		switch ia {
```
yaelharel, line 30: I think you should add a `default` case with a warning message, since theoretically someone can pass a value that isn't `Readable` nor `Writable` (because `ImageAccess` is an `int`).

testhelpers/docker_registry.go
```diff
@@ -182,6 +199,14 @@ func (r *DockerRegistry) EncodedLabeledAuth() string {
 	return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(`{"username":"%s","password":"%s"}`, r.username, r.password)))
 }
 
+// Add stores the given key name with the provided ImagePrivileges. For example
+// Add("my-image", NewImagePrivileges(Readable, Writable)) will save "my-image" as a
+// readable and writable image into the registry.
+// It must be called before Start to have any effect.
```
yaelharel, line 205: I don't think it's a good practice to ask the user to remember to call `Add` before calling `Start`.
It might be very confusing to find out the reason that your code isn't running the way you expect it to be because of such thing.
Maybe it's better to pass the `ImagePrivileges` to the `Start` function, or have two `Start` functions:
`StartWithPrivileges` and `StartWithoutPrivileges`.
These are just two ideas. Maybe there are better ones.

By the way, maybe I didn't follow but it looks like you're calling the `Add` function after the `Start` function in your [lifecycle PR](https://github.com/buildpacks/lifecycle/pull/685/files).

yaelharel (commented):
Hi @jjbustamante, these are great improvements.
Please see my comments.
Moreover, there are a few open discussions from your previous PR that are still relevant and I think we should consider continue discussing here (or open new issues for them) - https://github.com/buildpacks/imgutil/pull/125/files.
Thanks!

testhelpers/docker_registry.go
jjbustamante: @yaelharel , @natalieparellano.

The `Add` method can actually be called after the `Start`, as I mentioned before, it handler is holding a reference to the internal Map and every time the internal map is being updated after an `Add` invocation when the handler function is called the Map update is reflected.

I added a log entry in the Custom handler function to print the URL of the request and the content of the Map that is received, as you can see, the Map values are updated and the handler is called on every request made to the Docker registry on each test cases.

[local-out.txt](https://github.com/buildpacks/imgutil/files/6970269/local-out.txt)

testhelpers/docker_registry.go
```diff
@@ -68,17 +68,25 @@ func NewDockerRegistry(ops ...RegistryOption) *DockerRegistry {
...
+func BasicAuth(handler http.Handler, username, password, realm string, optPermissions ...map[string]ImagePrivileges) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if len(optPermissions) > 0 && (r.Method == "GET" || r.Method == "HEAD") {
+			permissions := optPermissions[0]
+			if permission, ok := permissions[extractImageName(r.URL.Path)]; ok {
+				if permission.readable {
+					handler.ServeHTTP(w, r)
+				}
+			}
+		}
```
natalieparellano, line 82: This sort of bothers me in that we're duplicating logic in `Custom()` (and also not handling the case when we want to "pass" a writeable image). It seems that what we want to do is IF the url is found in the map, just fall through to the regular handler (if permission is not denied). What if instead, we send `r.regHandler` into `Custom()` - and then around line 124, instead of falling through to the auth handler, we invoke `regHandler.Serve()`. Then we could delete this logic.

testhelpers/docker_registry.go
natalieparellano: Whoops, sorry, that was my code suggestion when I didn't understand how it worked!

natalieparellano:
@jjbustamante this is looking pretty good to me - LMK what you think of https://github.com/buildpacks/imgutil/pull/126/files#r687051113

testhelpers/image_privileges.go
```diff
@@ -25,16 +26,19 @@ const (
 // NewImagePrivileges(Readable) returns ImagePrivileges{readable: true, writable: false}
 // NewImagePrivileges(Writable) returns ImagePrivileges{readable: false, writable: true}
 // NewImagePrivileges(Readable, Writable) returns ImagePrivileges{readable: true, writable: true}
-func NewImagePrivileges(imageAccess ...ImageAccess) (priv ImagePrivileges) {
+func NewImagePrivileges(imageAccess ...ImageAccess) (ImagePrivileges) {
```
yaelharel, line 29: You don't need the `()` around the return type

testhelpers/docker_registry.go
jjbustamante: Keeping my idea of expressing the behavior with a diagram, doing this will work like this:

registry-3

The Custom handler decides which one to call based on the privileges and skipping the Basic Auth if it is a read operation. Yeap, I like it

testhelpers/docker_registry.go
jjbustamante: I deleted the comment in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1c2361a45515bb01cc7fe1fae051d620fdc643f3)

testhelpers/docker_registry.go
```diff
@@ -96,11 +94,13 @@ func ReadOnly(handler http.Handler) http.Handler {
...
 }
 
-func Custom(handler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
+func Custom(basicAuthHandler http.Handler, regHandler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if permission, ok := permissions[extractImageName(r.URL.Path)]; ok {
 			if r.Method == "GET" || r.Method == "HEAD" {
-				if !permission.readable {
+				if permission.readable {
+					regHandler.ServeHTTP(w, r)
```
natalieparellano, line 102: Wouldn't we also want to "wave through" a put request where the image has been explicitly configured to be writeable? Should we require basic auth in that case (I'd argue no). So there would be a similar extra lines right around line 109...

testhelpers/docker_registry.go
```diff
@@ -135,10 +140,14 @@ func (r *DockerRegistry) Start(t *testing.T) {
 	r.authnHandler = ReadOnly(r.regHandler)
 	if r.username != "" {
 		if r.customPrivileges != nil {
+			// wrap registry handler with basic auth
+			r.authnHandler = BasicAuth(r.regHandler, r.username, r.password, "registry")
 			// wrap registry handler with custom handler
-			r.regHandler = Custom(r.regHandler, r.customPrivileges)
+			r.authnHandler = Custom(r.authnHandler, r.regHandler, r.customPrivileges)
```
natalieparellano, line 146:
```suggestion
			r.authnHandler = Custom(basicAuthHandler, r.regHandler, r.customPrivileges)
```

natalieparellano: (I think this makes it a little more clear)

testhelpers/docker_registry.go
jjbustamante: If the image is writable, it will delegate the request to the Basic Auth handler that will check if you sent the credentials in the request, like the following image:

registry-3

jjbustamante: The idea of all this stuff, was because to implement the read-only behavior without forcing the client to send the Basic Auth value in the request, but the Write operation must validate the Auth, right? or you mean if the image is writable (by anyone) we can skip the validation? Not sure if we can do that in the registry right now, in that case, yes! we can add a similar logic in line 109

natalieparellano: That's interesting - I think you could make the case for it working either way. It's probably worth a quick discussion!

testhelpers/image_privileges.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/9fcb86825b0095e4dc69b5c128a8400944d50f04)

testhelpers/image_privileges.go
```diff
@@ -12,23 +12,36 @@ type ImagePrivileges struct {
...
-// NewImagePrivileges("foo-writable") returns ImagePrivileges{readable: false, writable: true}
-// NewImagePrivileges("foo-writable-readable") returns ImagePrivileges{readable: true, writable: true}
-func NewImagePrivileges(imageName string) (priv ImagePrivileges) {
-	priv.readable = strings.Contains(imageName, "readable")
-	priv.writable = strings.Contains(imageName, "writable")
+// NewImagePrivileges() returns ImagePrivileges{readable: false, writable: false}
+// NewImagePrivileges(Readable) returns ImagePrivileges{readable: true, writable: false}
+// NewImagePrivileges(Writable) returns ImagePrivileges{readable: false, writable: true}
+// NewImagePrivileges(Readable, Writable) returns ImagePrivileges{readable: true, writable: true}
+func NewImagePrivileges(imageAccess ...ImageAccess) (priv ImagePrivileges) {
```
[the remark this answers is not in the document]
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1c2361a45515bb01cc7fe1fae051d620fdc643f3)

testhelpers/image_privileges.go
jjbustamante: Removed

testhelpers/docker_registry.go
```diff
@@ -135,10 +140,14 @@ func (r *DockerRegistry) Start(t *testing.T) {
 	r.authnHandler = ReadOnly(r.regHandler)
 	if r.username != "" {
 		if r.customPrivileges != nil {
+			// wrap registry handler with basic auth
+			r.authnHandler = BasicAuth(r.regHandler, r.username, r.password, "registry")
```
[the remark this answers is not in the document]
jjbustamante: I agreed, applied in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/f19016e4dd378dcd0531ef1d351626730dd1ccb7)

testhelpers/docker_registry.go
jjbustamante: I agreed, applied in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/f19016e4dd378dcd0531ef1d351626730dd1ccb7)

testhelpers/docker_registry.go
natalieparellano: I think we might need a `return` after we call `regHandler.ServeHTTP()`. What do you think of the following?

```
var (
			image ImagePrivileges
			ok    bool
		)
		if image, ok = permissions[extractImageName(r.URL.Path)]; !ok {
			basicAuthHandler.ServeHTTP(w, r)
			return
		}
		if r.Method == "GET" || r.Method == "HEAD" {
			if !image.readable {
				w.WriteHeader(401)
				_, _ = w.Write([]byte("Unauthorized.\n"))
				return
			}
		} else { // assume write request
			if !image.writable {
				w.WriteHeader(401)
				_, _ = w.Write([]byte("Unauthorized.\n"))
				return
			}
		}
		regHandler.ServeHTTP(w, r)
```

testhelpers/docker_registry.go
```diff
@@ -193,11 +192,37 @@ func (r *DockerRegistry) EncodedLabeledAuth() string {
 	return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(`{"username":"%s","password":"%s"}`, r.username, r.password)))
 }
 
-// Add stores the given key name with the provided ImagePrivileges. For example
-// Add("my-image", NewImagePrivileges(Readable, Writable)) will save "my-image" as a
+// SetPrivilegesToImage sets the given image name with the provided ImagePrivileges if ImagePrivileges was enabled.
```
yaelharel, line 196: What do you mean by "if ImagePrivileges was enabled"?

testhelpers/docker_registry.go
```diff
@@ -182,6 +193,39 @@ func (r *DockerRegistry) EncodedLabeledAuth() string {
...
+// For example SetPrivilegesToImage("my-image", NewImagePrivileges(Readable, Writable)) will save "my-image" as a
+// readable and writable image into the registry.
+func (r *DockerRegistry) setPrivilegesToImage(imageName string, privilege ImagePrivileges) {
+	if r.isImagePrivilegesEnable() {
+		r.imagePrivileges[imageName] = privilege
+	}
+}
+
+// MakeReadOnly set the given image name to be readable when the ImagePrivileges feature was enabled
+func (r *DockerRegistry) MakeReadOnly(imageName string) {
```
yaelharel, line 206:
```suggestion
func (r *DockerRegistry) SetReadOnly(imageName string) {
```

yaelharel: If you agree with this suggestion, I would change the other functions as well.

testhelpers/docker_registry.go
```diff
@@ -96,29 +95,38 @@ func ReadOnly(handler http.Handler) http.Handler {
 	})
 }
 
-func Custom(handler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
+func delegator(basicAuthHandler http.Handler, regHandler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
```
yaelharel, line 98: I like the refactor in this function!

testhelpers/docker_registry.go
```diff
@@ -182,6 +193,39 @@ func (r *DockerRegistry) EncodedLabeledAuth() string {
...
+func (r *DockerRegistry) MakeReadWrite(imageName string) {
+	r.setPrivilegesToImage(imageName, NewImagePrivileges(Readable, Writable))
+}
+
+// MakeInaccessible set the given image name to do not have any access when the ImagePrivileges feature was enabled
+func (r *DockerRegistry) MakeInaccessible(imageName string) {
+	r.setPrivilegesToImage(imageName, NewImagePrivileges())
+}
+
+func (r *DockerRegistry) isImagePrivilegesEnable() bool {
```
yaelharel, line 225: I think that it should be: `isImagePrivilegesEnabled`

testhelpers/docker_registry.go
```diff
@@ -96,29 +95,38 @@ func ReadOnly(handler http.Handler) http.Handler {
 	})
 }
 
-func Custom(handler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
+func delegator(basicAuthHandler http.Handler, regHandler http.Handler, permissions map[string]ImagePrivileges) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
yaelharel, line 99: I think that I would change `r` and `w` to `req` and `res` or even to `request` and `response` (or maybe you have better names?).
While working on "read/write" permissions, it's a bit confusing that `w` doesn't influence the permissions because it's just the response writer.

testhelpers/docker_registry.go
natalieparellano: I like `imagePrivilegesEnabled`

testhelpers/docker_registry.go
```diff
@@ -182,6 +193,39 @@ func (r *DockerRegistry) EncodedLabeledAuth() string {
...
+// readable and writable image into the registry.
+func (r *DockerRegistry) setPrivilegesToImage(imageName string, privilege ImagePrivileges) {
+	if r.isImagePrivilegesEnable() {
+		r.imagePrivileges[imageName] = privilege
+	}
+}
+
+// MakeReadOnly set the given image name to be readable when the ImagePrivileges feature was enabled
+func (r *DockerRegistry) MakeReadOnly(imageName string) {
+	r.setPrivilegesToImage(imageName, NewImagePrivileges(Readable))
```
natalieparellano, line 207:
```suggestion
	r.setImagePrivileges(imageName, NewImagePrivileges(Readable))
```
(nit)

testhelpers/docker_registry.go
natalieparellano: This is pure syntactic sugar, but maybe you could make this function return the "repo name" for the image, so that [these lines](https://github.com/buildpacks/lifecycle/pull/685/files#diff-c4e4b4a55ebbd71443c5260020b287f3018cb7bb4c65c79e74554b7f97c45656R168-R169) could become 
```
fixtures.readOnlyRegCacheImage = testRegistry.MakeReadWrite(someCacheName) // or SetReadWrite

**Review by natalieparellano — APPROVED** [review]
@jjbustamante this is great! I agree with many of @yaelharel's suggestions for making the code more readable, but I also think we could merge it as-is (let us know what you'd like to do). It will be awesome to put these helpers to use in the lifecycle, this is going to simplify our tests so much! Thank you!

File: testhelpers/docker_registry.go (hunk 19 above)
Comment (jjbustamante) [reply testhelpers/docker_registry.go]: I like making the method to return the repoName. I added in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/c59c94c0966e2826f3e1ce6b48983be810952fee)

File: testhelpers/docker_registry.go [hunk 24]
```diff
@@ -96,29 +95,38 @@ func ReadOnly(handler http.Handler) http.Handler {
...
-				}
+		var (
+			image ImagePrivileges
+			ok    bool
+		)
+		if image, ok = permissions[extractImageName(r.URL.Path)]; !ok {
+			basicAuthHandler.ServeHTTP(w, r)
+			return
+		}
+		if r.Method == "GET" || r.Method == "HEAD" {
```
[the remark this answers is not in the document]
jjbustamante: Yes, it is also fine to add that method. I added in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/c59c94c0966e2826f3e1ce6b48983be810952fee)

testhelpers/docker_registry.go
jjbustamante: Renamed to `imagePrivilegesEnabled` in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/c59c94c0966e2826f3e1ce6b48983be810952fee)

testhelpers/docker_registry.go
jjbustamante: Cool, I renamed them in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/c59c94c0966e2826f3e1ce6b48983be810952fee)

testhelpers/docker_registry.go
jjbustamante: Fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/c59c94c0966e2826f3e1ce6b48983be810952fee)

testhelpers/docker_registry.go
jjbustamante: Well, because the method is only exposed in the package scope I added `if WithImagePrivileges was called during registry creation`. Is it clear?

testhelpers/docker_registry.go
```diff
@@ -71,54 +72,61 @@ func NewDockerRegistry(ops ...RegistryOption) *DockerRegistry {
...
-
 		handler.ServeHTTP(w, r)
 	})
 }
 
 // ReadOnly wraps a handler, allowing only GET and HEAD requests, otherwise rejecting with a 405
 func ReadOnly(handler http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if !(r.Method == "GET" || r.Method == "HEAD") {
+	return http.HandlerFunc(func(w http.ResponseWriter, request *http.Request) {
```
yaelharel, line 87:
```suggestion
	return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
```

yaelharel (approved):
Added a small suggestion but other than that, your PR looks great!
Thank you 😊

testhelpers/docker_registry.go
jjbustamante: Oh, I missed this one. Sorry it's fixed in the following [commit](https://github.com/buildpacks/imgutil/pull/126/commits/1eb46031019fbbfccbc8f307cb6d467508e41a7d)

PR #128, opened 2021-09-23
Better sparse image tar for podman

matejvasek:
This library uses the fact that uploaded tar may miss some layers when
pushed to docker provided the missing layers are already in daemon.
It works also with podman but the layers have to be present in the tar
at least as empty entries.

Signed-off-by: Matej Vasek <mvasek@redhat.com>

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -511,23 +511,30 @@ func (i *Image) doSave() (types.ImageInspect, error) {
 		return types.ImageInspect{}, err
 	}
 
+	var blankIdx int
 	var layerPaths []string
 	for _, path := range i.layerPaths {
 		if path == "" {
-			layerPaths = append(layerPaths, "")
-			continue
+			layerName := fmt.Sprintf("blank_%d", blankIdx)
+			blankIdx++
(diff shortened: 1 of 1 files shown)


matejvasek:
This make build faster and it also means fewer temporary files. This is especially important because there is leak a https://github.com/buildpacks/pack/issues/1167.

matejvasek:
this is based on findings of @vlk-charles

matejvasek:
for more reference https://github.com/containers/podman/issues/8132

matejvasek:
I'll check that.

matejvasek:
@natalieparellano I tried to run build on Java mvn project using `gcr.io/paketo-buildpacks/builder:base` and there seems to be no time performance drop.

natalieparellano (approved):
Looks good to me! Thank you @matejvasek for the contribution!

It would be nice to be able to test this, but that would require some reworking of our `doSave()` method. I think we could leave it off for now.

@jabrown85 what do you think?

jabrown85 (approved):

matejvasek:
@natalieparellano will you or somebody else update the dep in `pack` latter?

PR #137, opened 2022-02-03
Allow overriding image creation time

imjasonh:
If the `SOURCE_DATE_EPOCH` env var is set (to a number of seconds since the Unix epoch), its value will be used as the creation time of the image produced, and of config history values for each layer.

This matches the behavior of tools like [`ko`](https://github.com/google/ko#why-are-my-images-all-created-in-1970) and is the recommendation of https://reproducible-builds.org/

---

This was requested in a Slack thread, and I think unless there's a philosophical reason not to support it, the diff is small, and AFAIK will not have negative downstream effects (but isn't that always what they say?)

Anyway, let me know if there _is_ some philosophical reason not to support it, or if you suspect there will be negative downstream effects.


jabrown85 (approved):

natalieparellano (approved):

remote/remote.go
```diff
@@ -652,7 +668,7 @@ func (i *Image) Save(additionalNames ...string) error {
 	cfg.History = make([]v1.History, len(layers))
 	for i := range cfg.History {
 		cfg.History[i] = v1.History{
-			Created: v1.Time{Time: imgutil.NormalizedDateTime},
+			Created: creationTime,
```
natalieparellano, line 671: @ekcasey rightly [pointed out](https://buildpacks.slack.com/archives/CCP60GJLS/p1644533010840709?thread_ts=1644440749.308319&cid=CCP60GJLS) that we forgot about the daemon case - @imjasonh would you be amenable to adding that? It would go [here](https://github.com/buildpacks/imgutil/blob/14fe798ec25ae24f7225fd24cb11bf7a3e276a1d/local/local.go#L797).

edmorley:
I presume this can be closed, now that the alternative approach in #138 has merged?

PR #138, opened 2022-03-09
Allow callers to pass a custom CreatedAt value

jabrown85:
This will be used when resolving https://github.com/buildpacks/lifecycle/issues/809. The caller being able to adjust the value will allow us to read the value from different inputs if needed and won't tie imgutil to a specific env var.

Signed-off-by: Jesse Brown <jabrown85@gmail.com>

### DIFF
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -1607,7 +1607,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 			})
 
 			when("the WithCreatedAt option is used", func() {
-				it("zeroes times and client specific fields", func() {
+				it("uses the value for all times and client specific fields", func() {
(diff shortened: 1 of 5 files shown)


local/local_test.go
```diff
@@ -1606,6 +1606,38 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				}
 			})
 
+			when("the WithCreatedAt option is used", func() {
+				it("zeroes times and client specific fields", func() {
```
natalieparellano, line 1610:
```suggestion
				it("uses the provided time", func() {
```

natalieparellano (commented):
@jabrown85 looks great! Would it be possible also update the `fakes` package?

PR #142, opened 2022-04-26
Add empty layer before saving on remote images

jjbustamante:
# Issue

When a build is configured to use a gcr.io cache-image but, the underlying build does not produce any cache layers the cache writing fails with:

```bash
failed to write image to the following tags: [gcr.io/some-repo/some-cache-image:latest: PUT https://gcr.io/v2/some-repo/some-cache-image/manifests/latest: MANIFEST_INVALID: Failed to parse manifest for request "/v2/cf-build-service-dev-219913/test/cache-image/manifests/latest": Manifest is missing required field "layers".]
```

# Description
The following fix adds an option to the **remote.Image** implementation to add an Empty Layer just before saving the image into the registry to avoid the error describe above.

Fixes https://github.com/buildpacks/lifecycle/issues/830

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -2,6 +2,7 @@ package remote
 
 import (
 	"fmt"
+	"github.com/google/go-containerregistry/pkg/v1/static"
 	"io"
 	"net/http"
 	"strings"
@@ -31,6 +32,7 @@ type Image struct {
 	image      v1.Image
 	prevLayers []v1.Layer
 	createdAt  time.Time
+	addEmptyLayerOnSave bool
 }
 
 type options struct {
@@ -38,6 +40,7 @@ type options struct {
 	baseImageRepoName string
 	prevImageRepoName string
 	createdAt         time.Time
+	emptyLayer	  bool
 }
 
 type ImageOption func(*options) error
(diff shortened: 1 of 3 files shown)


remote/remote.go
```diff
@@ -2,6 +2,7 @@ package remote
 
 import (
 	"fmt"
+	"github.com/google/go-containerregistry/pkg/v1/static"
```
natalieparellano, line 5: nit: could you please group this line with the other `go-containerregistry` imports?

remote/remote.go
```diff
@@ -674,6 +688,11 @@ func (i *Image) Save(additionalNames ...string) error {
 	cfg = cfg.DeepCopy()
 
 	layers, err := i.image.Layers()
+	if len(layers) == 0 && i.addEmptyLayerOnSave {
+		empty := static.NewLayer([]byte{}, types.OCILayer)
+		i.image, err = mutate.AppendLayers(i.image, empty)
```
natalieparellano, line 693: Is it possible to derive the diffID of the layer? Could you call `GetLayer()` on it? That might make this testable (without having to modify the `Image` interface) per https://github.com/buildpacks/lifecycle/pull/852/files#r860932594

jjbustamante: In fact, that call throws the following error:  `mismatched fs layers (1) and diff ids (0)`

remote/remote.go
```diff
@@ -31,13 +32,15 @@ type Image struct {
...
 	createdAt  time.Time
+	addEmptyLayerOnSave bool
 }
 
 type options struct {
 	platform          imgutil.Platform
 	baseImageRepoName string
 	prevImageRepoName string
 	createdAt         time.Time
+	emptyLayer	  bool
```
natalieparellano, line 43: nit: is it possible we could name this variable `addEmptyLayerOnSave` as well?

natalieparellano:
@jjbustamante it looks good! I left a couple nits. Do you mind resolving the merge conflicts?

remote/remote.go
jjbustamante: Done

PR #144, opened 2022-04-29
tests: Add fake for `workingDir` inside `NewImage`

imnitishng:
Add fake for tests here https://github.com/buildpacks/pack/pull/1427


jromero:
@imnitishng Could you explain why this is necessary? Couldn't you use the already available `SetWorkingDir` to set a working dir? What's so special about the initialized value you provided?

imnitishng:
Ah yes, this totally slipped my mind, we can close this since I have updated the PR using `SetWorkingDir()`

PR #151, opened 2022-06-13
Narrow docker client interface

matejvasek:
Narrow docker client interface


matejvasek:
@natalieparellano once this is bumped in `pack` repo, I'll do similar changes there.

PR #154, opened 2022-08-01
Support custom access to docker image repository

wanjunlei:
Signed-off-by: wanjunlei <wanjunlei@kubesphere.io>

When creating a new imange, it can register a docker image registry with `WithRegistry` option like this

```
remote.NewImage(readOnlyRepoName, authn.DefaultKeychain).withRegistry("mydomain.registry.com:1080", true, false)
```

The first parameter is the URL of the docker image registry, the second is whether this registry is an insecure registry, and the third is whether to skip TLS authentication. And it can register serval registries.

This is useful when using a private docker image registry, and this pr is for the lifecycle to support a private docker image registry.

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -34,7 +34,7 @@ type Image struct {
 	prevLayers          []v1.Layer
 	createdAt           time.Time
 	addEmptyLayerOnSave bool
-	registries          []registry
+	registries          []reference
 }
 
 type options struct {
@@ -43,11 +43,11 @@ type options struct {
 	prevImageRepoName   string
 	createdAt           time.Time
 	addEmptyLayerOnSave bool
-	registries          []registry
+	registries          []reference
 }
 
-type registry struct {
-	repository         string
+type reference struct {
+	prefix             string
 	insecure           bool
 	insecureSkipVerify bool
 }
@@ -105,8 +105,8 @@ func AddEmptyLayerOnSave() ImageOption {
 // WithRegistry register a registry.
 func WithRegistry(repository string, insecure, insecureSkipVerify bool) ImageOption {
 	return func(opts *options) error {
-		opts.registries = append(opts.registries, registry{
-			repository:         repository,
+		opts.registries = append(opts.registries, reference{
+			prefix:             repository,
 			insecure:           insecure,
 			insecureSkipVerify: insecureSkipVerify,
 		})
@@ -142,15 +142,13 @@ func NewImage(repoName string, keychain authn.Keychain, ops ...ImageOption) (*Im
 	}
 
 	if imageOpts.prevImageRepoName != "" {
-		reg := getRegistry(imageOpts.prevImageRepoName, ri.registries)
-		if err := processPreviousImageOption(ri, imageOpts.prevImageRepoName, platform, reg); err != nil {
+		if err := processPreviousImageOption(ri, imageOpts.prevImageRepoName, platform); err != nil {
 			return nil, err
 		}
 	}
 
 	if imageOpts.baseImageRepoName != "" {
-		reg := getRegistry(imageOpts.prevImageRepoName, ri.registries)
-		if err := processBaseImageOption(ri, imageOpts.baseImageRepoName, platform, reg); err != nil {
+		if err := processBaseImageOption(ri, imageOpts.baseImageRepoName, platform); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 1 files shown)


remote/remote.go
```diff
@@ -41,6 +43,13 @@ type options struct {
 	prevImageRepoName   string
 	createdAt           time.Time
 	addEmptyLayerOnSave bool
+	registries          []registry
+}
+
+type registry struct {
+	repository         string
```
natalieparellano, line 50:
```suggestion
type reference struct {
	prefix         string
```
I wonder if this is more accurate, since `repository` as used in `getRegistry` could also contain a namespace or tag.

remote/remote.go
```diff
@@ -117,16 +138,19 @@ func NewImage(repoName string, keychain authn.Keychain, ops ...ImageOption) (*Im
 		repoName:            repoName,
 		image:               image,
 		addEmptyLayerOnSave: imageOpts.addEmptyLayerOnSave,
+		registries:          imageOpts.registries,
 	}
 
 	if imageOpts.prevImageRepoName != "" {
-		if err := processPreviousImageOption(ri, imageOpts.prevImageRepoName, platform); err != nil {
+		reg := getRegistry(imageOpts.prevImageRepoName, ri.registries)
```
natalieparellano, line 145: It might read better if this function call were inside `processPreviousImageOption`.

remote/remote.go
```diff
@@ -117,16 +138,19 @@ func NewImage(repoName string, keychain authn.Keychain, ops ...ImageOption) (*Im
...
+		if err := processPreviousImageOption(ri, imageOpts.prevImageRepoName, platform, reg); err != nil {
 			return nil, err
 		}
 	}
 
 	if imageOpts.baseImageRepoName != "" {
-		if err := processBaseImageOption(ri, imageOpts.baseImageRepoName, platform); err != nil {
+		reg := getRegistry(imageOpts.prevImageRepoName, ri.registries)
```
natalieparellano, line 152:
```suggestion
		reg := getRegistry(imageOpts.baseImageRepoName, ri.registries)
```
Is this what you meant?

remote/remote.go
```diff
@@ -265,9 +311,13 @@ func defaultPlatform() imgutil.Platform {
 	}
 }
 
-func referenceForRepoName(keychain authn.Keychain, ref string) (name.Reference, authn.Authenticator, error) {
+func referenceForRepoName(keychain authn.Keychain, ref string, reg *registry) (name.Reference, authn.Authenticator, error) {
```
natalieparellano, line 314:
```suggestion
func referenceForRepoName(keychain authn.Keychain, ref string, insecure bool) (name.Reference, authn.Authenticator, error) {
```

remote/remote.go
```diff
@@ -207,8 +241,8 @@ func prepareNewWindowsImage(ri *Image) error {
 	return nil
 }
 
-func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platform) (v1.Image, error) {
-	ref, auth, err := referenceForRepoName(keychain, repoName)
+func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platform, reg *registry) (v1.Image, error) {
```
natalieparellano, line 244:
```suggestion
func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platform, reg registry) (v1.Image, error) {
```
Does it need to be a pointer? Then you could avoid the nil check.

remote/remote.go
```diff
@@ -93,6 +102,18 @@ func AddEmptyLayerOnSave() ImageOption {
 	}
 }
 
+// WithRegistry register a registry.
+func WithRegistry(repository string, insecure, insecureSkipVerify bool) ImageOption {
```
natalieparellano, line 106: Would it be possible to add a test that calls `NewImage()` with this option?

natalieparellano (commented):
@wanjunlei thanks so much for the PR! In general this looks good to me. I added a couple comments and suggestions.

We've come across this problem before (see https://github.com/buildpacks/lifecycle/issues/524) and this would take us one step closer to solving it for other components 🎉

remote/remote.go
wanjunlei: I have no good idea to add a test for this option.
If it wants to verify the insecure option， it needs a docker registry with a non-localhost IP  or a non-private internets  IP, or a domain, this is not easy to achieve under the existing testing framework.

Do you have any suggestions?

natalieparellano, line 106:
```suggestion
// WithRegistrySetting registers options to use when accessing images in a registry in order to construct the image. The referenced images could include the base image, a previous image, or the image itself.
func WithRegistrySetting(repository string, insecure, insecureSkipVerify bool) ImageOption {
```

natalieparellano: I'm sorry if this is overly pedantic. It took me a bit of time to understand this code (why are there multiple registries per image? and so on...) and I feel like something like this would have helped.

remote/remote.go
```diff
@@ -207,8 +243,8 @@ func prepareNewWindowsImage(ri *Image) error {
 	return nil
 }
 
-func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platform) (v1.Image, error) {
-	ref, auth, err := referenceForRepoName(keychain, repoName)
+func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platform, reg reference) (v1.Image, error) {
```
natalieparellano, line 246:
```suggestion
func newV1Image(keychain authn.Keychain, repoName string, platform imgutil.Platform, reg registrySetting) (v1.Image, error) {
```

remote/remote.go
natalieparellano: Ah, you are right. Sigh... I suppose we could make the field on the Image struct public, but that feels unsatisfying. I'd be okay with not adding a test in this case.

natalieparellano (commented):
@wanjunlei thanks for updating! I left a couple more nits. But otherwise this seems good to go.

natalieparellano (approved):

natalieparellano:
Thanks for the PR @wanjunlei!

Issue #100, opened 2021-04-01
Support Saving images as archives.


dwillist:
# Summary
We would like to be able to call `img.SaveAsFile(filepath)` (or something similar) and have our images saved as some sort of archive. 

This ask has popped up in a couple of places:
buildpacks/pack#1125
buildpacks/lifecycle#423
buildpacks/pack#768 (needed for some of the exporter changes discussed here)

PR #155, opened 2022-08-17
Save local image to file as Docker archive

flix-:
\o/


jabrown85:
The new tests added [fail on windows](https://github.com/buildpacks/imgutil/runs/7978314675?check_suite_focus=true#step:4:432). The [specific error](https://github.com/buildpacks/imgutil/runs/7978314675?check_suite_focus=true#step:4:352) looks like `local_test.go:1814: Expected nil: invalid header name: must be absolute, posix path: foo`

flix-:
@jabrown85: Any chance I can manually trigger the test(s)? The latest failure looks quite odd and I don't have any Windows machine where I can test.

jabrown85:
I kicked it off

jabrown85:
@natalieparellano any ideas on how to make debugging this easier for someone who doesn't have access to a windows machine?

I see the error, but also don't have easy access to a windows machine to help.
```
TestLocal/Image/#SaveFile/previous_image_is_configured_and_layers_are_reused/creates_an_archive_that_can_be_imported_and_has_correct_diffIDs
    local_test.go:1797: Expected nil: Error: No such image: localhost:50061/pack-image-test-pehcgichfr
```

flix-:
@jabrown85: I am still not able to trigger the tests manually :/ Could you please retrigger the test?

flix-:
@jabrown85: :)

jjbustamante:
> @jabrown85: Please retrigger tests

@flix- I ran them ;)

flix-:
@jabrown85, @jjbustamante: Please re-rerun them again. If it fails again I need to check for some local Windows installation. This is wasting too much time :/

jabrown85:
Ugh, sorry. If you can't find the issue I'll try and look later this week on a windows machine.

jabrown85:
Hmm. I checked out your branch on my windows machine and it worked as expected. 😕 

Maybe this is a github action on windows issue?

```
Passed: 1 | Failed: 0 | Skipped: 67

=== CONT  TestLocal
    local_test.go:35: stop registry
--- PASS: TestLocal (1.56s)
    --- PASS: TestLocal/Image (1.52s)
        --- PASS: TestLocal/Image/#SaveFile/previous_image_is_configured_and_layers_are_reused/creates_an_archive_that_can_be_imported_and_has_correct_diffIDs (1.52s)
PASS
ok      github.com/buildpacks/imgutil/local     1.881s
```

natalieparellano:
@jabrown85 that would be sad indeed. I recall there were some Windows-specific oddities WRT writing tar records - see https://github.com/buildpacks/imgutil/blob/main/layer/windows_writer.go. Does the proposed implementation cover this logic? Perhaps some logic or methods could be borrowed from the `layer` package.

flix-:
@jabrown85, @natalieparellano: Thank you for your input. I have incorporated the windows-specific tar writer into the code. Lets give it another try. Although I am wondering a bit how the other tests without the windows-specifics went through :/

local/local.go
```diff
@@ -494,6 +501,116 @@ func (i *Image) Save(additionalNames ...string) error {
 	return nil
 }
 
+func (i *Image) SaveFile() (string, error) {
+	f, err := os.CreateTemp("", "imgutil.local.image.export.*.tar")
```
natalieparellano, line 505:
```suggestion
	if i.inspect.Os == "windows" {
		return "", errors.New("not yet implemented")
	}
	f, err := os.CreateTemp("", "imgutil.local.image.export.*.tar")
```

natalieparellano: WDYT @flix- @jabrown85? Do we need this to work for Windows images?

jabrown85: It does _technically_ work on my windows machine. Do we have a concept of an unsupported feature? It would be nice to allow it to work if it can and skip the tests instead as a way of saying "unsupported"

flix-: Managed to get a Windows machine... I assume it is working on your Windows clients because you are using Linux containers (right click on the Docker desktop tray icon and you'll see 'switch to Windows containers...'). IMHO Windows Server uses Windows containers per default. For Linux containers everything works as expected (osType is 'linux' and none of the Windows specifics apply). However for Windows containers I can reproduce the exact same error as reported by Github actions. I will need some time to investigate further. Current assumption is that the Windows base layers are the problematic part. If we have a Windows image they are added by default when using newImage(). I am reusing a layer here which also includes those base directories/files for windows containers (and a lot more than the ones generated by newImage()) and the image is imported without an error but is still not present afterwards. Manually removing the Windows base layer written by newImage() and only using the ones from the re-used layers works without any issues...

jabrown85: oh, doh! I did forget to change that. I think I assumed the tests would automate that bit for me. You can tell how often I use windows for container work 😬

flix-:
@jabrown85: Please re-run the test. Should work now -- At least its running on my Windows VM

flix-:
To give some background to the change: Windows is very picky with regards to the content of the layers. You need to have the registry and UtilityVM dirs/files as already specified in the base layer generation somewhere (does not need to be in a single layer though). Under certain conditions layers can be added without a 'Hives' (might be empty) directory. I have seen that when adding a UtilityVM directory only when it was not present in any layer before for example. If we assume we have those special directories already then it was always needed in the cases I have tested. The previous failing test had all those special directories already through the newImage() call and I tried to reuse the base layer from the nanoserver image which does not include the 'Hives' directory. Such a case is also failing with the image's Save() method by the way. It hast just not been tested in such a scenario.

In a nutshell: All images that can be succesfully imported via Save() can also be stored with SaveFile() now and re-imported via command line. Still there are cases where an image is saved to a file that cannot be imported. It would be very helpful to have a clear specification what is allowed or not in order to return an error fast.

There is one other thing I haven't understood: Why is it needed to have absolute paths in the tar archives for Windows? It was working without any issues with relative paths and also the images exported via command line on linux/windows didn't have such a constraint. By appending a '/' the tar files look quite ugly. I haven't touched it though to not break anything.

jjbustamante:
Does this PR fix, at least for a Local Image, this [issue](https://github.com/buildpacks/imgutil/issues/100)

flix-:
> Does this PR fix, at least for a Local Image, this [issue](https://github.com/buildpacks/imgutil/issues/100)

Yes!

natalieparellano:
> The previous failing test had all those special directories already through the newImage() call and I tried to reuse the base layer from the nanoserver image which does not include the 'Hives' directory

If I understand correctly, the fix involved changing the test to use `local.WithPreviousImage(prevImg.Name())` where the top layer **does** include the Hives directory, so that it can be re-used. Does that sound right?

> Why is it needed to have absolute paths in the tar archives for Windows?

I'm afraid this has been lost in the sands of time...

natalieparellano (approved):
Looks good to me... @jabrown85 would you like to take another look?

jabrown85 (approved):
Looks great! Thanks for sticking with it and hunting down that failing test.

PR #165, opened 2022-11-29
Implementation of CNB Image interface using OCI layout format

### LINKED ISSUE #162: Implement cnb.Image interface using OCI layout format
## Expected Behavior

New package `layout` will be exposed with the implementation of the `Image` [interface](https://github.com/buildpacks/imgutil/blob/3252b1f13dfa9ff71de9e82f86bd66ee7c598f48/image.go#L36) based on the [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) 

**Note**: There is a v1.Image wrapper [implementation](https://github.com/buildpacks/lifecycle/tree/main/internal/selective) in the lifecycle that mimics the behavior of a Image without actually having the layers on disk. 

The minimal implementation requires to expose the following classes:

``` mermaid
classDiagram
class ImageOption {
    <<interface>>
    FromBaseImage(base v1.Image) ImageOption
    FromBaseImagePath(path string) ImageOption
    WithPreviousImage(path string)
}
class LayoutImage {
    NewImage(path string, ops ...ImageOption) (*Image, error)
}
class SparseImage {
     NewImage(path string, from v1.Image) (*Image, error) 
}
```
Where
- LayoutImage: is actually *layout.Image* but mermaid doesn't allow to define the package name. Implements the `Image` [interface](https://github.com/buildpacks/imgutil/blob/3252b1f13dfa9ff71de9e82f86bd66ee7c598f48/image.go#L36) and handle the complexity of saving on Disk in OCI layout format.
- SparseImage: *sparse.Image* is an alternative implementation of the `Image` [interface](https://github.com/buildpacks/imgutil/blob/3252b1f13dfa9ff71de9e82f86bd66ee7c598f48/image.go#L36) that mimics the behavior onf a image without actually saving any layer on disk.

jjbustamante:
## Summary

This pull request implements the CNB Image interface using OCI layout format. 

A new `layout` package was added to expose the implementation of the `Image` interface

The implementation exposes the following classes:

``` mermaid
classDiagram
class ImageOption {
    <<interface>>
    FromBaseImage(base v1.Image) ImageOption
    FromBaseImagePath(path string) ImageOption
    WithPreviousImage(path string)
}
class LayoutImage {
    NewImage(path string, ops ...ImageOption) (*Image, error)
}
class SparseImage {
     NewImage(path string, from v1.Image) (*Image, error) 
}
```
Where
- LayoutImage: is actually *layout.Image* but mermaid doesn't allow to define the package name. Implements the `Image` interface and handle the complexity of saving on Disk in OCI layout format.
- SparseImage: *sparse.Image* is an alternative implementation of the `Image` interface that mimics the behavior of a image without actually saving any layer on disk. 

## Output

When the `image.Save()` method is called it will save the output in the local filesystem at the path provided during initialization in OCI layout format, for example: 

Screenshot 2023-01-11 at 8 15 51 PM

The previous image shows a couple of examples of two images save in OCI layout format on disk

#### Before

It was not possible to save images in OCI layout format

#### After

New package `layout` is exposed with the new implementation.

### Notes

- This is an experimental feature
- The following methods will not be implemented in this iteration
  - `image.Rebase()`
  - `image.SaveFile()`
  

Fixes #162 


layout/identifier.go
```diff
@@ -0,0 +1,28 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+type IDIdentifier struct {
+	ImageID string
+}
+
+func newIDIdentifier(configFile *v1.ConfigFile) (IDIdentifier, error) {
+	return IDIdentifier{
+		ImageID: "sha256:" + asSha256(configFile),
```
natalieparellano, line 16: I think elsewhere we call this "digest"

layout/layoutpath.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+}
+
+func WithAnnotations(annotations map[string]string) Option {
+	return func(i *options) {
+		i.annotations = annotations
+	}
+}
+
+func (l Path) path(elem ...string) string {
```
natalieparellano, line 32:
```suggestion
func (l Path) append(elem ...string) string {
```
This is maybe more descriptive

layout/sparse/sparse.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/buildpacks/imgutil/layout"
+
+	"github.com/buildpacks/imgutil"
+)
+
+var _ imgutil.Image = (*Image)(nil)
+
+type Image struct {
```
natalieparellano, line 15: We should add a docs comment to describe why this struct is needed

testhelpers/testhelpers.go
```diff
@@ -391,3 +399,133 @@ func checkResponseError(r io.Reader) error {
...
+	AssertNil(t, err)
+
+	return configFile
+}
+
+func ReadManifestAndConfigFile(t *testing.T, path string) (*v1.Manifest, *v1.ConfigFile) {
+	index := ReadIndexManifest(t, path)
+	AssertEq(t, len(index.Manifests), 1)
+
+	// TODO add platform to select the Manifest
```
natalieparellano, line 527: Do we still need this?

testhelpers/testhelpers.go
```diff
@@ -391,3 +399,133 @@ func checkResponseError(r io.Reader) error {
 
 	return nil
 }
+
+func CreateSingleFileTar(path, txt string) (io.Reader, error) {
```
natalieparellano, line 403:
```suggestion
func CreateSingleFileTarReader(path, txt string) (io.Reader, error) {
```
This might be better for distinguishing this function from `CreateSingleFileLayerTar`

testhelpers/testhelpers.go
```diff
@@ -391,3 +399,133 @@ func checkResponseError(r io.Reader) error {
...
+
+	_, err = io.Copy(mw, r)
+	AssertNil(t, err)
+
+	sha = hex.EncodeToString(hasher.Sum(make([]byte, 0, hasher.Size())))
+
+	return path, "sha256:" + sha, contentsBuf.Bytes()
+}
+
+func CreateRemoteImage(t *testing.T) v1.Image {
```
natalieparellano, line 441:
```suggestion
func RunnableRemoteImage(t *testing.T) v1.Image {
```
We might change `RunnableBaseImage` to `RunnableBaseImageReference` to associate these methods

layout/identifier.go
```diff
@@ -0,0 +1,28 @@
+package layout
+
+import (
+	"crypto/sha256"
+	"fmt"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+type IDIdentifier struct {
```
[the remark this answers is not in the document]
natalieparellano: Or maybe `Digest`? I'm actually a bit confused why we need this.

layout/write.go
```diff
@@ -0,0 +1,263 @@
...
+
+	"github.com/google/go-containerregistry/pkg/logs"
+	"github.com/google/go-containerregistry/pkg/v1/stream"
+	"golang.org/x/sync/errgroup"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
+)
+
+// AppendImage mimics GGCR's `sparse` AppendImage in that it appends an underlyingImage to a `layout.Path`,
```
natalieparellano, line 20:
```suggestion
// AppendImage mimics GGCR's AppendImage in that it appends an underlyingImage to a `layout.Path`,
```

layout/write.go
```diff
@@ -0,0 +1,263 @@
...
+// but the underlyingImage appended does not include any layers in the `blobs` directory.
+// The returned underlyingImage will return layers when Layers(), LayerByDiffID(), or LayerByDigest() are called,
+// but the returned layer will error when DiffID(), Compressed(), or Uncompressed() are called.
+// This is useful when we need to satisfy the v1.Image interface but do not need to access any layers, such as when extending
+// base images with kaniko.
```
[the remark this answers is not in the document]
natalieparellano: `the returned layer will error when DiffID(), Compressed(), or Uncompressed() are called` - is this still a thing?

layout/sparse/sparse_test.go
```diff
@@ -0,0 +1,93 @@
...
+				err = image.Save("my-additional-tag")
+				h.AssertNil(t, err)
+
+				//  expected blobs: manifest, config, layer
+				h.AssertBlobsLen(t, imagePath, 2)
+
+				// assert additional name
+				index := h.ReadIndexManifest(t, imagePath)
+				h.AssertEq(t, len(index.Manifests), 1)
+				h.AssertEq(t, "my-additional-tag", index.Manifests[0].Annotations["org.opencontainers.image.ref.name"])
```
natalieparellano, line 66: Do we need this behavior? We don't do this for remote images...

layout/sparse/sparse_test.go
```diff
@@ -0,0 +1,93 @@
...
+		it.Before(func() {
+			imagePath = filepath.Join(tmpDir, "sparse-layout-image")
+		})
+
+		it.After(func() {
+			// removes all images created
+			os.RemoveAll(imagePath)
+		})
+
+		when("additional names are provided", func() {
```
natalieparellano, line 51:
```suggestion
		when("name(s) provided", func() {
```

layout/sparse/sparse_test.go
```diff
@@ -0,0 +1,93 @@
...
+				//  expected blobs: manifest, config, layer
+				h.AssertBlobsLen(t, imagePath, 2)
+
+				// assert additional name
+				index := h.ReadIndexManifest(t, imagePath)
+				h.AssertEq(t, len(index.Manifests), 1)
+				h.AssertEq(t, "my-additional-tag", index.Manifests[0].Annotations["org.opencontainers.image.ref.name"])
+			})
+
+			it("failed on saved when more than one name is provided", func() {
```
natalieparellano, line 69: I wonder if instead of erroring we could just ignore them? Or enforce that they are always blank?

layout/layout.go
```diff
@@ -0,0 +1,740 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/pkg/errors"
+
+	"github.com/buildpacks/imgutil"
+)
+
+var _ imgutil.Image = (*Image)(nil)
+
+type Image struct {
```
natalieparellano, line 23: We should add some comment that Image extends a v1.Image

layout/layout.go
```diff
@@ -0,0 +1,740 @@
...
+	return imgutil.Platform{
+		OS:           "linux",
+		Architecture: "amd64",
+	}
+}
+
+func (i *Image) Label(key string) (string, error) {
+	cfg, err := i.Image.ConfigFile()
+	if err != nil {
+		return "", errors.Wrapf(err, "getting config file for image at path %q", i.path)
```
natalieparellano, line 183:
```suggestion
		return "", fmt.Errorf(err, "getting config for image at path %q: %w", i.path, err)
```
(to be consistent with following error and remove use of deprecated `github.com/pkg/errors`)

natalieparellano: Since this sort of error handling is used a lot in this file, maybe we could make a helper for it

layout/layout.go
```diff
@@ -0,0 +1,740 @@
...
+	}
+	return cfg.Architecture, nil
+}
+
+func (i *Image) Name() string {
+	return i.path
+}
+
+func (i *Image) Rename(name string) {
+	i.path = name
```
natalieparellano, line 286: If `Rename` is called and the image is saved, does it copy/move the bits?

layout/layout.go
```diff
@@ -0,0 +1,740 @@
...
+func (i *Image) CreatedAt() (time.Time, error) {
+	configFile, err := i.Image.ConfigFile()
+	if err != nil {
+		return time.Time{}, errors.Wrapf(err, "getting createdAt time for image at path %q", i.path)
+	}
+	return configFile.Created.UTC(), nil
+}
+
+func (i *Image) Rebase(s string, image imgutil.Image) error {
+	return nil
```
natalieparellano, line 314: Let's return a not-implemented error instead, that way callers can be alerted

layout/layout.go
```diff
@@ -0,0 +1,740 @@
...
+	if len(diagnostics) > 0 {
+		return imgutil.SaveError{Errors: diagnostics}
+	}
+
+	return nil
+}
+
+func (i *Image) SaveFile() (string, error) {
+	return "", errors.New("not yet implemented")
+}
```
natalieparellano, line 575: What are the differences in output here? Between a local image `SaveFile()` and a layout image `Save()`?

layout/layout.go
```diff
@@ -0,0 +1,740 @@
...
+		}
+	} else {
+		image, err = emptyImage(platform)
+		if err != nil {
+			return nil, errors.Wrap(err, "initializing empty image")
+		}
+	}
+	return &Image{
+		Image: image,
+		path:  path, // TODO: is this needed?
```
natalieparellano, line 666:
```suggestion
		path:  path,
```

layout/layout_test.go
```diff
@@ -0,0 +1,486 @@
...
+			os.RemoveAll(imagePath)
+		})
+
+		when("#FromBaseImage", func() {
+			it.Before(func() {
+				imagePath = filepath.Join(tmpDir, "save-from-base-image")
+			})
+
+			when("additional names are provided", func() {
+				it("creates an image with all the layers from the underlying image and org.opencontainers.image.ref.name annotation", func() {
```
natalieparellano, line 65: As mentioned elsewhere, maybe we can reduce complexity by not caring about additional names, unless there is a reason to care.

layout/layout_test.go
```diff
@@ -0,0 +1,486 @@
...
+		// removes all images created
+		os.RemoveAll(tmpDir)
+	})
+
+	when("#Save", func() {
+		it.After(func() {
+			os.RemoveAll(imagePath)
+		})
+
+		when("#FromBaseImage", func() {
```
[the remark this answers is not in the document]
natalieparellano: Do we still need both `FromBaseImage` and `FromBaseImagePath`? Since the remote and local packages only have one helper, it might introduce confusion if layout package has this special case.

layout/layout_test.go
```diff
@@ -0,0 +1,486 @@
...
+						// assert additional name
+						index := h.ReadIndexManifest(t, imagePath)
+						h.AssertEq(t, len(index.Manifests), 1)
+						h.AssertEq(t, "v1.0", index.Manifests[0].Annotations["org.opencontainers.image.ref.name"])
+					})
+				})
+			})
+		})
+
+		when("#FromPreviousImage", func() {
```
natalieparellano, line 165: What if the previous image is sparse?

layout/layout_test.go
```diff
@@ -0,0 +1,486 @@
...
+			os.RemoveAll(imagePath)
+		})
+
+		it.Before(func() {
+			imagePath = filepath.Join(tmpDir, "feature-image")
+			image, err = layout.NewImage(imagePath)
+			h.AssertNil(t, err)
+		})
+
+		when("#Label(s)", func() {
```
natalieparellano, line 262:
```suggestion
		when("#Labels", func() {
```

layout/layout_test.go
```diff
@@ -0,0 +1,486 @@
...
+				h.AssertNil(t, err)
+
+				_, configFile = h.ReadManifestAndConfigFile(t, imagePath)
+
+				labels = configFile.Config.Labels
+				h.AssertEq(t, len(labels), 0)
+			})
+		})
+
+		when("#Env(s)", func() {
```
natalieparellano, line 303:
```suggestion
		when("#Env", func() {
```

layout/layout_test.go
```diff
@@ -0,0 +1,486 @@
...
+
+				it("returns true", func() {
+					h.AssertTrue(t, func() bool {
+						return image.Found()
+					})
+				})
+			})
+		})
+
+		when("#CMD", func() {
```
natalieparellano, line 410:
```suggestion
		when("#SetCmd", func() {
```

natalieparellano:
@jjbustamante added a few nits... I had some questions in my previous set of comments that would be good to answer, but overall I'm feeling really good about this PR :)

layout/layout_test.go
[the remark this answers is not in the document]
jjbustamante: Well, so far the use cases for both are:

- Pack: when pack prepares the inputs for the lifecycle, `FromBaseImage` seems to be a good option to save on disk the `run-image` on disk
- Lifecycle: because the `layout` feature assumes all the inputs references are on disk, then it will use `FromBaseImagePath` to load the images

layout/layout.go
natalieparellano: That sounds about right to me

layout/layout_test.go
jjbustamante: I think I missed that test case, we can't reuse the layer if it doesn't exist from the previous image (because it is sparse) so I think we should err but not sure.

testhelpers/testhelpers.go
jjbustamante: There is already a method with that name `CreateSingleFileTarReader`.  This one is creating a tar file with its header

layout/sparse/sparse_test.go
jjbustamante: Maybe a warn message could be a good solution

testhelpers/testhelpers.go
jjbustamante: Well, for testing I do not think so

layout/identifier.go
jjbustamante: Yeah, I took the same name used in remote package

layout/layout_test.go
jjbustamante: Well, it seems to be working, the layers are added into the Manifest but they are not present in the `blobs`

layout/sparse/sparse_test.go
jjbustamante: I removed this, and just added a message to the log to warn the user

layout/sparse/sparse_test.go
jjbustamante: Done

layout/layout.go
jjbustamante: - If you call the rename `before` saving, then the image is going to be saved in the new destination
- If you call the rename `after` saving and then call the `save` again, then the image is going to be saved in the new path, but the original one is not deleted or anything. Maybe it could be smarter but I do not want to complicate the implementation right now.

layout/layout.go
jjbustamante: Done

layout/layout_test.go
```diff
@@ -0,0 +1,486 @@
...
+
+				// from testdata/layout/busybox-sparse/
+				diffID := "sha256:40cf597a9181e86497f4121c604f9f0ab208950a98ca21db883f26b0a548a2eb"
+				_, err = image.GetLayer(diffID)
+				h.AssertNil(t, err)
+			})
+		})
+	})
+
+	when("#Feature", func() {
```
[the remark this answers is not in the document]
jjbustamante: Done, I refactored the `layout_test.go` file and know it has the same structure than the `remote_test.go` file

jjbustamante:
Hi @jabrown85 

If you can take a look and provide some feedback will be awesome! I already went to some of the comments Natalie left me.

layout/sparse/sparse.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil, err
+	}
+
+	image := &Image{
+		Image: *img,
+	}
+	return image, nil
+}
+
+func (i *Image) Save(additionalNames ...string) error {
```
jabrown85, line 35: I think implementing `SaveAs` is simple and probably worth doing now so we don't run into any issues using the `Image` interface. This goes for `layout.Image` as well. Rebase is going to use `SaveAs` once [this pr](https://github.com/buildpacks/lifecycle/pull/985) lands.

```suggestion
func (i *Image) Save(additionalNames ...string) error {
	return i.SaveAs(i.Name(), additionalNames...)
}

func (i *Image) SaveAs(name string, additionalNames ...string) error {
```

jabrown85 (changes requested):
One simple change requested. This is look great to me! Great work.

layout/sparse/sparse.go
jjbustamante: Yeah, I did this change, my mistake the method I want to leave non-implemented is **SaveFile()**

jabrown85 (approved):
This looks good!

PR #175, opened 2023-01-27
Adding capability to write the org.opencontainers.image.ref.name to OCI layout image

### LINKED ISSUE #174: Add annotation `org.opencontainers.image.ref.name` when image in OCI layout is saved
## Description

As part of the [discussion](https://github.com/buildpacks/rfcs/pull/203#discussion_r1088225327) of the export to OCI layout [RFC](https://github.com/buildpacks/rfcs/pull/203) we require to add the  `org.opencontainers.image.ref.name` annotation to the final OCI layout image when it is saved.

The reason to that is because other tools in the ecosystem uses or add that annotation. For example: 
 - [umoci](https://umo.ci/) is a tool used to create a runtime bundle from an image in OCI layout format, it requires the [annotation](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) `org.opencontainers.image.ref.name` to be present. 
 - [skopeo](https://github.com/containers/skopeo) tool include the annotation when an image is `copy` in OCI layout format.

## Acceptance Criteria

When an image is created using the `layout.NewImage(...)` method the resulting image on disk after `Save()`, `SaveAs()` is called should contain the `org.opencontainers.image.ref.name` annotation as described in the [spec](https://specs.opencontainers.org/image-spec/annotations/?v=v1.0.1)

jjbustamante:
## Description

This PR is adding the capability to the Layout image to add the `org.opencontainers.image.ref.name` annotation when the image is saved on disk.

Also, I included a modification to the `Image.Identifier()` method according to the discussion in this [thread](https://cloud-native.slack.com/archives/C0331B5QS02/p1674761476209019) 

### Additional change included

As I mentioned before, I've being trying to use [umoci](https://umo.ci/) to validate the output image generated by `layout` package. I noticed, that this tools only recognized OCI media types, like these:

```bash
Manifest - application/vnd.oci.image.manifest.v1+json
Config - application/vnd.oci.image.config.v1+json
Layers - application/vnd.oci.image.layer.v1.tar+gzip
```
Apparently, ggcr uses docker media types by default.

```bash
Manifest - application/vnd.docker.distribution.manifest.v2+json
Config - application/vnd.docker.container.image.v1+json
Layers - application/vnd.docker.image.rootfs.diff.tar.gzip
```
The original implementation was writing the image in OCI layout format using docker media types, as a consequence, if I tried to create a runtime bundle using  [umoci](https://umo.ci/) I was getting errors like:

```bash
 ⨯ create runtime bundle: unpack rootfs: unpack rootfs: layer 
sha256:72d9f18d70f395ff9bfae4d193077ccea3ca583e3da3dd66f5c84520c0100727: 
blob is not correct mediatype: application/vnd.docker.image.rootfs.diff.tar.gzip
```
 or something similar complaining about `application/vnd.docker.distribution.manifest.v2+json`

This PR also included some logic to workaround this, the idea is to validate if the given base image (path or v1.Image) has docker media types in such as case, I am trying to create a new image updating the media types. ggcr offers `mutate.MediaTypes` methods to do that BUT I had a lot of troubles trying to change the Layers media types.

Fixes #174 

### DIFF
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -291,6 +291,14 @@ func (i *Image) Found() bool {
 	return !i.deleted
 }
 
+func (i *Image) AnnotateRefName(refName string) error {
+	return nil
+}
+
+func (i *Image) GetAnnotateRefName() (string, error) {
+	return "", nil
+}
+
 // test methods
 
 func (i *Image) SetIdentifier(identifier imgutil.Identifier) {
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -33,6 +33,7 @@ func NewImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
 		os:            "linux",
 		osVersion:     "",
 		architecture:  "amd64",
+		annotations:   map[string]string{},
 	}
 }
 
@@ -58,6 +59,8 @@ type Image struct {
 	workingDir    string
 	savedNames    map[string]bool
 	manifestSize  int64
+	refName       string
+	annotations   map[string]string
 }
 
 func (i *Image) CreatedAt() (time.Time, error) {
@@ -240,6 +243,9 @@ func (i *Image) SaveAs(name string, additionalNames ...string) error {
 	}
 
 	allNames := append([]string{name}, additionalNames...)
+	if i.refName != "" {
+		i.annotations["org.opencontainers.image.ref.name"] = i.refName
+	}
 
 	var errs []imgutil.SaveDiagnostic
 	for _, n := range allNames {
@@ -292,11 +298,12 @@ func (i *Image) Found() bool {
 }
 
 func (i *Image) AnnotateRefName(refName string) error {
+	i.refName = refName
 	return nil
 }
 
 func (i *Image) GetAnnotateRefName() (string, error) {
-	return "", nil
+	return i.refName, nil
 }
 
 // test methods
@@ -440,4 +447,8 @@ func (i *Image) SetManifestSize(size int64) {
 
 func (i *Image) ManifestSize() (int64, error) {
 	return i.manifestSize, nil
+}
+
+func (i *Image) Annotations() map[string]string {
+	return i.annotations
 }
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -20,47 +20,47 @@ import (
 
 func NewImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
 	return &Image{
-		labels:        nil,
-		env:           map[string]string{},
-		topLayerSha:   topLayerSha,
-		identifier:    identifier,
-		name:          name,
-		cmd:           []string{"initialCMD"},
-		layersMap:     map[string]string{},
-		prevLayersMap: map[string]string{},
-		createdAt:     time.Now(),
-		savedNames:    map[string]bool{},
-		os:            "linux",
-		osVersion:     "",
-		architecture:  "amd64",
-		annotations:   map[string]string{},
+		labels:           nil,
+		env:              map[string]string{},
+		topLayerSha:      topLayerSha,
+		identifier:       identifier,
+		name:             name,
+		cmd:              []string{"initialCMD"},
+		layersMap:        map[string]string{},
+		prevLayersMap:    map[string]string{},
+		createdAt:        time.Now(),
+		savedNames:       map[string]bool{},
+		os:               "linux",
+		osVersion:        "",
+		architecture:     "amd64",
+		savedAnnotations: map[string]string{},
 	}
 }
 
 type Image struct {
-	deleted       bool
-	layers        []string
-	layersMap     map[string]string
-	prevLayersMap map[string]string
-	reusedLayers  []string
-	labels        map[string]string
-	env           map[string]string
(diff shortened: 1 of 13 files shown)


jabrown85 (approved):

jjbustamante:
@jabrown85, @natalieparellano if you can take a look at this PR will be great! 

I am actually fixing two things here, I can create another issue and link it to this one, but at least I tried to explain and provide context in the description, please let me know your thoughts

layout/layout.go
```diff
@@ -733,11 +730,33 @@ func (i *Image) mutateCreatedAt(base v1.Image, created v1.Time) error {
 
 // mutateImage wraps the provided v1.Image into a layout.Image
 func (i *Image) mutateImage(base v1.Image) {
-	i.Image = &Image{
-		Image: base,
+	manifest, _ := base.Manifest()
```
natalieparellano, line 733: Is there a reason not to check the error?

layout/layout.go
```diff
@@ -733,11 +730,33 @@ func (i *Image) mutateCreatedAt(base v1.Image, created v1.Time) error {
...
+	} else {
+		// images has docker media types, we need to override them
+		newBaseImage := overridesMediaType(base)
+		i.Image = &Image{
+			Image: newBaseImage,
+		}
 	}
 }
 
+// addOCILayer append the provided layer with media type application/vnd.oci.image.layer.v1.tar+gzip
```
natalieparellano, line 747:
```suggestion
// addOCILayer appends the provided layer with media type application/vnd.oci.image.layer.v1.tar+gzip
```

layout/layout.go
```diff
@@ -747,3 +766,40 @@ func imageRefAnnotation(additionalNames ...string) map[string]string {
...
+// validMediaTypes returns true if media types present in the manifest are the ones defined by the OCI spec
+// Docker Media Types will return false.
+func validMediaTypes(manifest *v1.Manifest) bool {
+	return manifest.MediaType == types.OCIManifestSchema1 &&
+		manifest.Config.MediaType == types.OCIConfigJSON
+}
+
+// overridesMediaType will create a new v1.Image from the provided base image, but replacing
+// manifest media type, config media type and layers media type by the ones defined by the OCI spec
+func overridesMediaType(base v1.Image) v1.Image {
```
natalieparellano, line 779:
```suggestion
func overrideMediaTypes(base v1.Image) v1.Image {
```

layout/layout.go
```diff
@@ -747,3 +766,40 @@ func imageRefAnnotation(additionalNames ...string) map[string]string {
...
+	config, _ := base.ConfigFile()
```
natalieparellano, line 780: Should we be checking errors here?

layout/layout.go
```diff
@@ -559,7 +549,8 @@ func (i *Image) SaveAs(name string, additionalNames ...string) error {
 	}
 
 	var diagnostics []imgutil.SaveDiagnostic
-	err = path.AppendImage(i.Image)
+	annotations := imageRefAnnotation(additionalNames...)
```
natalieparellano, line 552: It's a bit confusing that we ignore the originally provided `name` here... are we still in the situation where `name` is the path and `additionalNames` are the annotations?

layout/layout.go
jjbustamante: Same as my above comment, I will change the method signature to throw the error.

layout/layout.go
jjbustamante: Yeah, we were, but based on our talk I just changed the logic

layout/layout.go
jjbustamante: Done

fakes/image.go
```diff
@@ -291,6 +291,14 @@ func (i *Image) Found() bool {
...
 }
 
+func (i *Image) AnnotateRefName(refName string) error {
+	return nil
+}
+
+func (i *Image) GetAnnotateRefName() (string, error) {
+	return "", nil
+}
+
```
natalieparellano, line 301: Should we be updating the fake somehow? To avoid confusing future consumers of this fake...

remote/remote.go
```diff
@@ -865,6 +865,12 @@ func (i *Image) ManifestSize() (int64, error) {
 	return i.image.Size()
 }
 
+func (i *Image) AnnotateRefName(refName string) error { return nil }
```
natalieparellano, line 868: We could throw a not implemented error, since in theory you might want to add this annotation for a remote image. It would be nice to know that it doesn't work

natalieparellano (approved):
I made a couple small suggestions but in general this looks great! Amazing work @jjbustamante!

natalieparellano:
@jabrown85 did you want to take another look?

jabrown85:
@natalieparellano you all go ahead without me on this one!

fakes/image.go
jjbustamante: Well, I added an implementation for those methods, the annotation is being added during `Save()` (if it was provided) and a test method `Annotations()` was exposed, that will help users of this implementation to test some behaviors.

natalieparellano (approved):

PR #184, opened 2023-03-08
Image: add Variant, SetVariant to interface

joe-kimmel-vmw:
required to complete https://github.com/buildpacks/lifecycle/issues/742 , because now we'll be setting the architecture variant by reading from the RunImage.


jjbustamante (commented):
LGTM.
I will just suggest to add the test coverage

PR #186, opened 2023-03-22
Refactor: file organization

natalieparellano:
This might be a tad OCD, but I sat down to work on imgutil and I found myself a bit confused as to why we have certain functions i.e., what they are used for, and wanting to compare implementations across remote/local/layout and having to dig around to find each one. This PR:

* organizes the methods on the interface
* orders the interface implementations by how the interface methods are declared (grouping getters, setters, etc)
* creates new files for "new" and "save" operations in each package as this is where we have most of the custom logic
* a couple other small changes (variable rename, etc) to make things consistent across packages

Hopefully this will make it easier for folks new to imgutil to find stuff and know what things are used for.

I didn't touch the test files so that we can be confident that nothing broke, but eventually we might want to apply some organization there as well.


jjbustamante (approved):
I think the refactoring makes a lot of sense, and actually it makes the code more easy to read for new contributors.

PR #187, opened 2023-03-22
Feature: override media types for layout and remote

natalieparellano:
This will help in a couple of cases:
* We can delete the [internal/selective](https://github.com/buildpacks/lifecycle/tree/main/internal/selective) package in the lifecycle and use layout/sparse instead (with media type option)
* If we desire to continue publishing lifecycle images that have docker media types (now that `gcr.io/distroless/static` has updated), we can do so

This builds off of refactor/file-organization so branching off of that until that one is merged

### DIFF
--- a/image.go
+++ b/image.go
@@ -5,33 +5,12 @@ import (
 	"io"
 	"strings"
 	"time"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/types"
 )
-
-var NormalizedDateTime = time.Date(1980, time.January, 1, 0, 0, 1, 0, time.UTC)
-
-type SaveDiagnostic struct {
-	ImageName string
-	Cause     error
-}
-
-type SaveError struct {
-	Errors []SaveDiagnostic
-}
-
-func (e SaveError) Error() string {
-	var errors []string
-	for _, d := range e.Errors {
-		errors = append(errors, fmt.Sprintf("[%s: %s]", d.ImageName, d.Cause.Error()))
-	}
-	return fmt.Sprintf("failed to write image to the following tags: %s", strings.Join(errors, ","))
-}
-
-// Platform represents the target arch/os/os_version for an image construction and querying.
-type Platform struct {
-	Architecture string
-	OS           string
-	OSVersion    string
-}
 
 type Image interface {
 	// getters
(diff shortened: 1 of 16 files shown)


image.go
```diff
@@ -5,33 +5,12 @@ import (
...
-	Errors []SaveDiagnostic
-}
-
-func (e SaveError) Error() string {
-	var errors []string
-	for _, d := range e.Errors {
-		errors = append(errors, fmt.Sprintf("[%s: %s]", d.ImageName, d.Cause.Error()))
-	}
-	return fmt.Sprintf("failed to write image to the following tags: %s", strings.Join(errors, ","))
-}
```
natalieparellano, line 27: Moved later in the file, it probably makes sense for the interface declaration to be first

layout/layout.go
```diff
@@ -492,82 +492,24 @@ func findLayerWithSha(layers []v1.Layer, diffID string) (v1.Layer, error) {
...
-		newBaseImage, err := overrideMediaTypes(base)
-		if err != nil {
-			return err
-		}
-		i.Image = &Image{
-			Image: newBaseImage,
-		}
-	}
-	return nil
-}
```
natalieparellano, line 536: Moved to layout/new.go and renamed `setUnderlyingImage`. We may want to make a separate function that checks and updates media types, as we don't really need to do that on the level of the image after it is first instantiated.

remote/new.go
```diff
@@ -76,6 +76,13 @@ func NewImage(repoName string, keychain authn.Keychain, ops ...ImageOption) (*Im
 		ri.createdAt = imageOpts.createdAt
 	}
 
+	if imageOpts.mediaTypes != "" {
+		ri.requestedMediaTypes = imageOpts.mediaTypes
+	}
```
natalieparellano, line 81:
```suggestion
	ri.requestedMediaTypes = imageOpts.mediaTypes
```
This accomplishes the same thing

image.go
```diff
@@ -90,3 +69,109 @@ type Image interface {
...
+func layersAddendum(layers []v1.Layer, mediaType types.MediaType) []mutate.Addendum {
+	additions := make([]mutate.Addendum, 0)
+	for _, layer := range layers {
+		additions = append(additions, mutate.Addendum{
+			MediaType: mediaType,
+			Layer:     layer,
+		})
+	}
+	return additions
+}
```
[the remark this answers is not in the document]
jjbustamante: Nice ...

jjbustamante (approved):
It looks really good to me ❤️

layout/new.go
```diff
@@ -57,6 +57,15 @@ func NewImage(path string, ops ...ImageOption) (*Image, error) {
 		ri.createdAt = imageOpts.createdAt
 	}
 
+	if imageOpts.mediaTypes != "" {
+		ri.requestedMediaTypes = imageOpts.mediaTypes
+	} else {
+		ri.requestedMediaTypes = imgutil.OCITypes // without media types option, default to oci media type
```
natalieparellano, line 63: Actually, thinking about how this would work in the lifecycle, I do think we want an option that is like "I don't care, just use whatever media types are already there" ...so might update `imgutil.DefaultTypes` to be `default` instead of `""`.

natalieparellano: Updated in eaefdf2

jabrown85 (approved):

PR #190, opened 2023-04-03
add Validate function

matthewrobertson:
This adds a new `Validate` function that can be used to validate image contents in a registry. For remote images, `Validate` invokes go-containerregistry's validate function with the `Fast` option enabled, so it verifies the existence of layers via HTTP HEAD requests. For layout and local images, `Validate` is just an alias for `Found`.

We will use this function in the lifecycle to address https://github.com/buildpacks/lifecycle/issues/1044

### DIFF
--- a/fakes/image.go
+++ b/fakes/image.go
@@ -307,6 +307,10 @@ func (i *Image) Found() bool {
 	return !i.deleted
 }
 
+func (i *Image) Valid() bool {
+	return !i.deleted
+}
+
 func (i *Image) AnnotateRefName(refName string) error {
 	i.refName = refName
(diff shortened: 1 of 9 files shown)


jabrown85 (approved):

natalieparellano (approved):

natalieparellano:
@matthewrobertson should we pull this one in?

matthewrobertson:
@natalieparellano yes I think it is good to do. I will also need to make a small change in lifecycle to start using this new function. But I am not sure how this project gets consumed by lifecycle... Once it this is merged, can I just use it?

PR #191, opened 2023-04-05
Manifest List Interface

### LINKED ISSUE #188: Create an new interface to handle Manifest Lists
## Context

The OCI spec defines an [Image Index ](https://github.com/opencontainers/image-spec/blob/main/image-index.md) concept to handle multiple manifests in an OCI image. Currently, we expose an [Image](https://github.com/buildpacks/imgutil/blob/main/image.go) interface that is consumed no matter the implementation (daemon, registry or layout), this is very convenient to decouple the implementation. 

## Requeriment

This issue suggests the definition and implementation of a new *interface* to expose the operations to deal with an [Image Index ](https://github.com/opencontainers/image-spec/blob/main/image-index.md).

## Proposal

Note: I am using underscore to refer to a package, for example, foo_Bar referes to a *Bar* class in the *foo* package

```mermaid
classDiagram
    ManifestList <|-- remote_ManifestList
    ManifestList <|--  local_ManifestList 
    remote_ManifestList o-- ManifestListOption

    class ManifestListOption {
        +WithMediaTypes(requested imgutil.MediaTypes)
        +WithPath(path string)
    }

    class remote_ManifestList {
        +NewManifestList(repoName string, keychain authn.Keychain, ops []ManifestListOption) ManifestList 
    }

    class local_ManifestList {
         +NewManifestList(repoName string, path string, ops []ManifestListOption) ManifestList
    }

    class ManifestList {
       <<interface>>
        +Add(repoName string) error
        +Remove(repoName string) error
        +Save() error
    }
```

## Notes

- *remote_ManifestList* must be added into the `imgutil.remote` package in a `manifest_list.go` file

husni-faiz:
Hi all,

I am working on the issue #188 to create a ManifestList/ImageIndex interface as part of a larger scope of adding multi-architecture support to buildpacks. I would appreciate your ideas and thoughts on how we want to move forward. There is a lot of validations and test to be added along the line.

In parallel a pack-cli command `manifest` is being developed to work with Manifest Lists.

You may find the following resources useful.

- [Multi-Architecutre goals](https://docs.google.com/document/d/1nrMmz8ZFOdEwzlzQYS963BfSqCvfoyhbk5FvExueuM4/edit#heading=h.2bfpr0eysfzq)
- [ImageIndex](https://github.com/opencontainers/image-spec/blob/main/image-index.md)
- [Meeting minutes](https://docs.google.com/document/d/169eIH-1tYZi9TC1ZZH32OmkOc0F1PTyu-6Dwcn30YnU/edit?usp=sharing)
- [Slack Discussion](https://cloud-native.slack.com/archives/C0331B61A1Y/p1680196805987539)
- pack [manifest](https://github.com/buildpacks/pack/pull/1705) command

Fixes #188


jjbustamante (commented):
Keep working on this

local/index.go
```diff
@@ -0,0 +1,249 @@
...
+
+	cfg, err := img.ConfigFile()
+
+	if err != nil {
+		return errors.Wrapf(err, "getting config file for image %q", repoName)
+	}
+	if cfg == nil {
+		return fmt.Errorf("missing config for image %q", repoName)
+	}
+	if cfg.OS == "" {
```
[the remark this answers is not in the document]
drac98: I have removed the code validating `OS` and `Arch` and moved the validation to pack-cli so that the client won't allow pushing an index with incomplete platform information.

local/index.go
```diff
@@ -0,0 +1,249 @@
...
+	return nil
+}
+
+func makeFileSafeName(ref string) string {
+	fileName := strings.Replace(ref, ":", "-", -1)
+	return strings.Replace(fileName, "/", "_", -1)
+}
+
+func createManifestListDirectory(transaction string) error {
+	path := filepath.Join("/home/drac98/.docker/manifests", makeFileSafeName(transaction))
```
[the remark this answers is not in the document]
drac98: pack-cli will use the config.PackHome() path as the default path to store local indexes.

WYGIN:
Can you please update this branch, @husni-faiz?

layout/save.go
```diff
@@ -27,6 +28,51 @@ func (i *Image) SaveAs(name string, additionalNames ...string) error {
...
+		if osVersion, _ := i.OSVersion(); osVersion != "" {
+			config.Platform.OSVersion = osVersion
+		}
+
+		mfest.Config = config
+	})
+	if err != nil {
+		return err
+	}
+
```
jjbustamante, line 75: @WYGIN actually these lines are causing the timeout when the Layout tests are executed, if I remove it, current tests are fine

WYGIN: The issue is raised due to concurrency, I just noticed it and fixed it.

fakes/image.go
```diff
@@ -37,32 +43,276 @@ func NewImage(name, topLayerSha string, identifier imgutil.Identifier) *Image {
...
+	var stringSlice []string
+	for key, value := range data {
+		keyValue := fmt.Sprintf("%s=%s", key, value)
+		stringSlice = append(stringSlice, keyValue)
+	}
+	return stringSlice
+}
+
+// ConfigFile implements v1.Image.
+func (i *Image) ConfigFile() (*v1.ConfigFile, error) {
```
jjbustamante, line 87: This doesn't look like a **Fake** implementation, looks like we want to duplicate the real logic and it doesn't make a lot of sense.

fakes/image.go
```diff
@@ -501,3 +801,48 @@ func (i *Image) ManifestSize() (int64, error) {
 func (i *Image) SavedAnnotations() map[string]string {
 	return i.savedAnnotations
 }
+
+// uncompressedLayer implements partial.UncompressedLayer from raw bytes.
+type uncompressedLayer struct {
```
jjbustamante, line 806: Similar this is not a **Fake** implementation

remote/save.go
```diff
@@ -16,72 +15,50 @@ func (i *Image) Save(additionalNames ...string) error {
...
+
+		os, _ := i.OS()
+		arch, _ := i.Architecture()
+		variant, _ := i.Variant()
+		osv, _ := i.OSVersion()
+		features, _ := i.Features()
+		osFeatures, _ := i.OSFeatures()
+		urls, _ := i.URLs()
+		annotations, _ := i.Annotations()
+
```
WYGIN, line 51: I believe this should be 
```go
                os, _ := i.os
		arch, _ := i.arch
		variant, _ := i.variant
		osv, _ := i.osVersion
		features, _ := i.features
		osFeatures, _ := i.osFeatures
		urls, _ := i.urls
		annotations, _ := i.annotations
```
as these fields will keep track of updated values, calling `i.OS()` might not return the value of `i.os`, 

- `i.OS()` will return the OS of the image that is before the user modified
- `i.os` will return the updated OS the user wants the image to be saved as.

for example, a user wants an Image with `linux/amd64` to be updated to `windows/amd64`
the `i.OS()` still returns `linux`, whereas `i.os` returns `windows`

local/new.go
```diff
@@ -64,6 +69,52 @@ func NewImage(repoName string, dockerClient DockerClient, ops ...imgutil.ImageOp
...
+		return idx, err
+	}
+
+	return &ImageIndex{
+		CNBIndex: cnbIndex,
+	}, nil
+}
+
 func defaultPlatform(dockerClient DockerClient) (imgutil.Platform, error) {
 	daemonInfo, err := dockerClient.ServerVersion(context.Background())
```
WYGIN, line 119: since we have moved to GGCR, do we still need to consider using Docker Client?

index/new.go
```diff
@@ -0,0 +1,50 @@
...
+
+	var cnbIndex *imgutil.CNBIndex
+	switch idxOps.Format {
+	case types.DockerManifestList:
+		cnbIndex, err = imgutil.NewCNBIndex(imgutil.NewEmptyDockerIndex(), *idxOps)
+		if err != nil {
+			return idx, err
+		}
+		// TODO I don't think we should write into disk during creation
+		_, err = layout.Write(layoutPath, imgutil.NewEmptyDockerIndex())
```
jjbustamante, line 36: I already started removing this logic, I did it for the layout package, I will do the same for local and remote

layout/new_test.go
```diff
@@ -0,0 +1,237 @@
...
+						h.AssertEq(t, annotations["org.opencontainers.image.url"], "https://hub.docker.com/_/busybox")
+						h.AssertEq(t, annotations["org.opencontainers.image.revision"], "185a3f7f21c307b15ef99b7088b228f004ff5f11")
+					})
+				})
+			})
+
+			when("#URLs", func() {
+				when("linux/amd64", func() {
+					it("existing annotations are readable", func() {
+						t.Skip("Do we really want to support this now???, its failing")
```
jjbustamante, line 228: Maybe I didn't set the `subject` data correctly, but I didn't know how to test this. Maybe we can remove that method, honestly, I do not thing the community is asking for it and it adds more complexity to the code

local/new.go
jjbustamante: We still talk with the daemon, so yes we still need the docker client. Natalie is working on isolating the code for the interaction with the daemon and keeping all the common stuff around the v1.Image for the local, remote and layout package.

layout/new_test.go
WYGIN: I personally felt the same, But i saw similar implementation in podman to Set and Get URLs, I think it is using those setters and getters internally for something else.

Currently, the setters like `SetFeatures`, `SetOSFeatures`, `SetAnnotations` actually append data instead of replacing(similar to Podman and Docker), I wonder **what if the CNB author wants to remove any feature, osFeature or annotation**?

At present I think it is not a priority, incase if its helpful to CNB authors I propose adding `features`, `osFeatures` and `annotations` flags to `pack manifest rm` command

- when `pack manifest rm < some-image | some-index >`, the image or index with given digest will be removed
- when `pack manifest rm < some-image | some-index > [ --features | --os-features | --annotations ]` then instead of removing image, pack removes features, osFeatures and annotations from specified image.

index/new.go
WYGIN: Actually, the [empty.Index](https://github.com/google/go-containerregistry/blob/051d642e295e3e86b58bb5feabb184036fa34a52/pkg/v1/empty/index.go#L51-L57) has a couple of methods that always returns an error that we often tend to call, So I thought writing it to dist and loading with [layout](https://github.com/google/go-containerregistry/blob/051d642e295e3e86b58bb5feabb184036fa34a52/pkg/v1/layout/index.go#L84-L123) might work like a hack to fix these

util.go
```diff
@@ -0,0 +1,253 @@
...
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/partial"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/pkg/errors"
+)
+
+func GetConfigFile(image v1.Image) (*v1.ConfigFile, error) {
```
natalieparellano, line 16: Nice :)

util.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, err
+	}
+	if manifest == nil {
+		return nil, errors.New("missing manifest")
+	}
+	return manifest, nil
+}
+
+func MutateManifest(i v1.Image, withFunc func(c *v1.Manifest) (mutateSubject, mutateAnnoations bool)) (v1.Image, error) {
+	// FIXME: put MutateManifest on the interface when `remote` and `layout` packages also support it.
```
natalieparellano, line 39: Each of remote, layout, and local should implement the v1.Image interface now. However the local manifest is going to contain garbage data like empty diffIDs and layer sizes of -1. We should probably detect where this is going to blow up and handle errors gracefully.

cnb_image.go
```diff
@@ -26,6 +26,8 @@ type CNBImageCore struct {
 	preferredMediaTypes MediaTypes
 	preserveHistory     bool
 	previousImage       v1.Image
+	features            []string
+	annotations         map[string]string
```
natalieparellano, line 30: Why do these have to be fields on the struct? Could they not live on the manifest and config?

image.go
```diff
@@ -7,12 +7,42 @@ import (
 	"time"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
+type EditableImage interface {
```
natalieparellano, line 13: Nice organization :) I'm curious to know why the other SetXXX methods (e.g., SetCmd) aren't included. And are methods like CreatedAt and Entrypoint not also "getters"?

local/store.go
```diff
@@ -360,6 +360,25 @@ func (s *Store) SaveFile(image *Image, withName string) (string, error) {
 		return "", err
 	}
 
+	image.Image, err = imgutil.MutateManifest(image.Image, func(mfest *v1.Manifest) (mutateSubject, mutateAnnotations bool) {
```
natalieparellano, line 363: Even though a local.Image implements the v1.Image interface, the tarball generated from SaveFile is in the legacy docker format (not OCI layout) so I don't think this does anything.

local/local.go
```diff
@@ -22,6 +23,7 @@ type Image struct {
 	store          *Store
 	lastIdentifier string
 	daemonOS       string
+	mutex          sync.Mutex
```
natalieparellano, line 26: Why do we need the mutex? What's the race condition here?

natalieparellano (commented):
@WYGIN @jjbustamante @husni-faiz thanks for this :) I'm still going through it, but had a few questions/comments in the meantime

local/local.go
WYGIN: We are accessing [index manifest](https://github.com/buildpacks/imgutil/blob/6eb1a7027489a6508012fd461b27a5bd065e3780/local/store.go#L364) pointer concurrently, anyhow we are removing the MutateManifest due to image saving in legacy docker format, this is no more needed.

image.go
WYGIN: The intention behind `EditableImage` interface is to create a type that has Setters an Getters that we need to annotate an image using `pack manifest` command
And [Here](https://github.com/buildpacks/imgutil/blob/6eb1a7027489a6508012fd461b27a5bd065e3780/options.go#L112) we can actually add a local/layout image that should have these set of methods must be defined, and for [MultiArch builder](https://github.com/buildpacks/pack/blob/a78dd269bb8bd5bf566fca7ebba72120430baee0/pkg/buildpack/builder.go#L86) I am using this interface to mutate image accordingly

cnb_image.go
WYGIN: these fields exists in [v1.Manifest](https://github.com/google/go-containerregistry/blob/051d642e295e3e86b58bb5feabb184036fa34a52/pkg/v1/manifest.go#L25-L53) but not in [v1.ConfigFile](https://github.com/google/go-containerregistry/blob/051d642e295e3e86b58bb5feabb184036fa34a52/pkg/v1/config.go#L29-L42), 
If these fields should be removed then, I think we should change the code as following, so that mutating `os`, `arch`, `variant`, `osVersion`, `osFeatures` should reflect changes in both ConfigFile and v1.Manifest
```go
	func (i *CNBImageCore) MutateConfigFile(withFunc func(c *v1.ConfigFile)) error {
	// FIXME: put MutateConfigFile on the interface when `remote` and `layout` packages also support it.
	configFile, err := getConfigFile(i.Image)
	if err != nil {
		return err
	}
	withFunc(configFile)
	i.Image, err = mutate.ConfigFile(i.Image, configFile)
	return MutateManifest(.....)
}
```

natalieparellano: I am not sure that you need to explicitly call MutateManifest - the image that you get back from calling `mutate.ConfigFile` is going to re-compute the manifest when you call `image.Manifest()` or whatever. Unless I'm misunderstanding something. I don't see `arch`, `variant`, etc. in the [manifest spec](https://github.com/opencontainers/image-spec/blob/main/manifest.md) so I assume you mean re-computing the config digest in the manifest which should be covered.

local/local.go
natalieparellano: Hmm... I'm still confused. When we're adding an image to a manifest list it's the index that's being mutated, not the image. So shouldn't the mutex live on the index?

cnb_image.go
natalieparellano: So looking at GGCR, I think we are good - the index will compute the descriptor from the added image; as long as the image has the right platform, this will bubble up to the index: https://github.com/google/go-containerregistry/blob/051d642e295e3e86b58bb5feabb184036fa34a52/pkg/v1/mutate/index.go#L123

WYGIN: I was not aware of [manifest spec](https://github.com/opencontainers/image-spec/blob/main/manifest.md) not having `os`, `arch`, `variant` etc.., field 

I think we can now safely delete `MutateManifest` calls for `CNBImage` but at the same time we should also change getters and setters of `CNBImage` `Features` to process `mfest.Subject` instead of `mfest.Config` or should we return nil?

And I also think that this change will also cleanup a lot of code from this PR(as I thought to add `os`, `arch`, `variant` which are not getting added by GGCR which significantly increased PR code size)

local/local.go
WYGIN: I believe we no longer need this, we can completely remove `MutateManifest` calls, which caused race condition here :(

cnb_image.go
natalieparellano: I am not sure that we need to worry about manifest.Subject for the time being. Looking at the spec:

> subject [descriptor](https://github.com/opencontainers/image-spec/blob/main/descriptor.md)

> This OPTIONAL property specifies a [descriptor](https://github.com/opencontainers/image-spec/blob/main/descriptor.md) of another manifest. This value defines a weak association to a separate [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree) structure, and is used by the [referrers API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers) to include this manifest in the list of responses for the subject digest.

I believe this is when the index in question is an aggregate of SBOMs or signatures relating to another image. Since we are just publishing regular images we shouldn't need it.

layout/layout_test.go
```diff
@@ -1174,21 +1083,42 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			h.AssertEq(t, configFile.OSVersion, osVersion)
+			h.AssertEq(t, configFile.OSFeatures, osFeatures)
+
+			h.AssertEq(t, mfest.Subject.Platform.OS, os)
+			h.AssertEq(t, mfest.Subject.Platform.Architecture, arch)
+			h.AssertEq(t, mfest.Subject.Platform.Variant, variant)
+			h.AssertEq(t, mfest.Subject.Platform.OSVersion, osVersion)
+			h.AssertEq(t, mfest.Subject.Platform.Features, features)
+			h.AssertEq(t, mfest.Subject.Platform.OSFeatures, osFeatures)
+			h.AssertEq(t, mfest.Subject.Annotations, annos)
```
natalieparellano, line 1119:
```suggestion
			h.AssertEq(t, mfest.Config.Platform.OS, os)
			h.AssertEq(t, mfest.Config.Platform.Architecture, arch)
			h.AssertEq(t, mfest.Config.Platform.Variant, variant)
			h.AssertEq(t, mfest.Config.Platform.OSVersion, osVersion)
			h.AssertEq(t, mfest.Config.Platform.Features, features)
			h.AssertEq(t, mfest.Config.Platform.OSFeatures, osFeatures)
			h.AssertEq(t, mfest.Config.Annotations, annos)
```

layout/layout_test.go
```diff
@@ -1224,3 +1147,431 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+						// manifest was added
+						// initially it has 2 manifest + 1 new
+						h.AssertEq(t, len(index.Manifests), 3)
+					})
+				})
+			})
+		})
+	})
+
+	when("#Push", func() {
```
natalieparellano, line 1442: Should this go in the remote package?

layout/layout_test.go
```diff
@@ -1224,3 +1147,431 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 		})
 	})
 }
+
+func testImageIndex(t *testing.T, when spec.G, it spec.S) {
```
natalieparellano, line 1151: Organizational nit: should this go in its own file?

cnb_image.go
natalieparellano: Will be resolved by https://github.com/husni-faiz/imgutil/pull/5

index.go
```diff
@@ -0,0 +1,31 @@
...
+	OSVersion(digest name.Digest) (osVersion string, err error)
+	Variant(digest name.Digest) (osVariant string, err error)
+
+	// setters
+
+	SetAnnotations(digest name.Digest, annotations map[string]string) error
+
+	// misc
+
+	Add(repoName string, ops ...func(options *IndexAddOptions) error) error
```
natalieparellano, line 25: Shouldn't we do something like `Add(image v1.Image, ...) and make the caller responsible for instantiating the image? That would make this work across layout/remote etc.

natalieparellano: @jjbustamante WDYT?

remote/new_test.go
```diff
@@ -0,0 +1,137 @@
+package remote_test
```
natalieparellano, line 1: Same comment as https://github.com/buildpacks/imgutil/pull/191/files?show-viewed-files=true&file-filters%5B%5D=#r1569540596

cnb_index.go
```diff
@@ -0,0 +1,1254 @@
...
+	RepoName         string
+	XdgPath          string
+	annotate         Annotate
+	KeyChain         authn.Keychain
+	Format           types.MediaType
+	removedManifests []v1.Hash
+	images           map[v1.Hash]v1.Descriptor
+}
+
+// Annotate a helper struct used for keeping track of changes made to ImageIndex.
```
natalieparellano, line 73: @WYGIN is it possible to provide a little more context on this struct and why it's needed?

cnb_index.go
```diff
@@ -0,0 +1,334 @@
...
+// AddManifest adds an image to the index.
+func (h *CNBIndex) AddManifest(image v1.Image) {
+	h.ImageIndex = mutate.AppendManifests(h.ImageIndex, mutate.IndexAddendum{
+		Add: image,
+	})
+}
+
+// SaveDir will locally save the index.
+func (h *CNBIndex) SaveDir() error {
+	layoutPath := filepath.Join(h.XdgPath, MakeFileSafeName(h.RepoName)) // FIXME: do we create an OCI-layout compatible directory structure?
```
WYGIN, line 207: yes

cnb_index.go
```diff
@@ -0,0 +1,381 @@
...
+		remote.WithTransport(GetTransport(pushOps.Insecure)),
+	)
+	if err != nil {
+		return err
+	}
+
+	if pushOps.Purge {
+		return h.DeleteDir()
+	}
+	return h.SaveDir()
```
WYGIN, line 340: isn't the `--purge` option used to delete local imageIndex after the pushing is successful?

cnb_image.go
```diff
@@ -285,6 +312,12 @@ func (i *CNBImageCore) SetOS(osVal string) error {
 	})
 }
 
+func (i *CNBImageCore) SetOSFeatures(osFeatures []string) error {
+	return i.MutateConfigFile(func(c *v1.ConfigFile) {
+		c.OSFeatures = osFeatures
```
WYGIN, line 317: isn't the expected behaviour is to append the requested `osFeatures` instead of replacing/overriding with existing one?

options.go
```diff
@@ -93,3 +98,136 @@ func WithPreviousImage(name string) func(*ImageOptions) {
...
+// WithXDGRuntimePath Saves the Index to the '`xdgPath`/manifests'
+func WithXDGRuntimePath(xdgPath string) func(options *IndexOptions) error {
+	return func(o *IndexOptions) error {
+		o.XdgPath = xdgPath
+		return nil
+	}
+}
+
+// PullInsecure If true, pulls images from insecure registry
+func PullInsecure() func(options *IndexOptions) error {
```
natalieparellano, line 169: Could we use `remote.WithRegistrySetting` instead?

options.go
```diff
@@ -93,3 +98,136 @@ func WithPreviousImage(name string) func(*ImageOptions) {
...
+// Push the Index with given format
+func WithTags(tags ...string) func(options *IndexPushOptions) error {
+	return func(a *IndexPushOptions) error {
+		a.Tags = tags
+		return nil
+	}
+}
+
+// Push index to Insecure Registry
+func WithInsecure(insecure bool) func(options *IndexPushOptions) error {
```
natalieparellano, line 205: Do we need to have separate options for push and pull?

options.go
```diff
@@ -93,3 +98,136 @@ func WithPreviousImage(name string) func(*ImageOptions) {
...
+	return func(a *IndexPushOptions) error {
+		a.Insecure = insecure
+		return nil
+	}
+}
+
+// Push the Index with given format
+func UsingFormat(format types.MediaType) func(options *IndexPushOptions) error {
```
natalieparellano, line 213: Could we just set the media type when the object is instantiated, and have one media type option?

remote/new.go
```diff
@@ -175,3 +219,21 @@ func NewV1Image(baseImageRepoName string, keychain authn.Keychain, ops ...func(*
 	options.Platform = processPlatformOption(options.Platform)
 	return processImageOption(baseImageRepoName, keychain, options.Platform, options.RegistrySettings)
 }
+
+// ValidateRepoName
+// TODO move this code to something more generic
+func validateRepoName(repoName string, o *imgutil.IndexOptions) error {
```
natalieparellano, line 225: This function will be more re-usable if it just accepts an insecure bool, instead of index options. Then we can use it for images too.

layout/new.go
```diff
@@ -58,6 +63,67 @@ func NewImage(path string, ops ...imgutil.ImageOption) (*Image, error) {
...
+	}
+
+	if idxOps.BaseIndex == nil && idxOps.BaseImageIndexRepoName != "" {
+		idxOps.BaseIndex, err = newImageIndexFromPath(idxOps.BaseImageIndexRepoName)
+		if err != nil {
+			return idx, err
+		}
+
+		if idxOps.BaseIndex != nil {
+			// TODO Do we need to do this?
```
natalieparellano, line 87: Ehh probably not

layout/new.go
```diff
@@ -58,6 +63,67 @@ func NewImage(path string, ops ...imgutil.ImageOption) (*Image, error) {
...
+	}
+
+	if idxOps.BaseIndex == nil {
+		switch idxOps.Format {
+		case types.DockerManifestList:
+			idxOps.BaseIndex = imgutil.NewEmptyDockerIndex()
+		default:
+			idxOps.BaseIndex = empty.Index
+		}
+	}
```
natalieparellano, line 114: This logic can probably go in the CNBIndex constructor

new.go
```diff
@@ -291,3 +291,20 @@ func prepareNewWindowsImageIfNeeded(image *CNBImageCore) error {
 	}
 	return nil
 }
+
+func NewCNBIndex(repoName string, v1Index v1.ImageIndex, ops IndexOptions) (*CNBIndex, error) {
```
natalieparellano, line 295: Why are we providing `v1Index` as a separate argument when it is contained in `ops`?

layout/new.go
```diff
@@ -58,6 +63,67 @@ func NewImage(path string, ops ...imgutil.ImageOption) (*Image, error) {
 	}, nil
 }
 
+// NewIndex will return an OCI ImageIndex saved on disk using OCI media Types. It can be modified and saved to a registry
+func NewIndex(repoName, path string, ops ...imgutil.IndexOption) (idx *ImageIndex, err error) {
```
natalieparellano, line 67: The only option that we actually use in this function is `idxOps.BaseImageIndexRepoName`. What if this function just returned a `v1.ImageIndex` (instead of wrapping it in `CNBIndex`)? That way the caller could use this to construct an index to provide to `imgutil.NewCNBIndex`.

remote/new.go
```diff
@@ -60,6 +60,50 @@ func NewImage(repoName string, keychain authn.Keychain, ops ...imgutil.ImageOpti
 	}, nil
 }
 
+// NewIndex returns a new ImageIndex from the registry that can be modified and saved to local file system
+func NewIndex(repoName string, ops ...imgutil.IndexOption) (idx *ImageIndex, err error) {
```
natalieparellano, line 64: Similar comment to https://github.com/buildpacks/imgutil/pull/191/files#r1572576794, I don't think this function needs to know about a CNBIndex.

cnb_index.go
natalieparellano: Yeah, but we return on line 338 so if we purge we never get to line 340

options.go
natalieparellano: Resolved by https://github.com/husni-faiz/imgutil/pull/11

options.go
```diff
@@ -93,3 +98,136 @@ func WithPreviousImage(name string) func(*ImageOptions) {
...
+// PullInsecure If true, pulls images from insecure registry
+func PullInsecure() func(options *IndexOptions) error {
+	return func(o *IndexOptions) error {
+		o.Insecure = true
+		return nil
+	}
+}
+
+// WithFormat Create the image index with the following format
+func WithFormat(format types.MediaType) func(options *IndexOptions) error {
```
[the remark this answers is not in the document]
natalieparellano: Resolved by https://github.com/husni-faiz/imgutil/pull/11

options.go
natalieparellano: Resolved by https://github.com/husni-faiz/imgutil/pull/11

cnb_image.go
natalieparellano: @jjbustamante do you know?

natalieparellano: Maybe we should call it `AppendOSFeatures`? But also, are we using this method in pack?

layout/new.go
jjbustamante: I think we should removes it, we try to load an index from path and according to the option `FromBaseImageIndex`, **If the index is not found, it does nothing** and we are throwing an error

layout/new.go
jjbustamante: Makes sense, apparently we don't have any particular behavior for `layout` or `remote` package so wrapping the `CNBIndex` is not adding any value right now.

layout/new_test.go
```diff
@@ -0,0 +1,106 @@
+package layout_test
```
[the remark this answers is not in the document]
jjbustamante: Yes, it reflects better its purpose

new.go
jjbustamante: I think at some point I wanted to make mandatory that if you want to create a CNBIndex you need a v1.ImageIndex during creation, but we do [this](https://github.com/buildpacks/imgutil/pull/191/files/e9716dbb3f7abae4ff8986ac44cb0a147c4cf815#r1575089124) then we don't need it anymore

remote/new.go
jjbustamante: I agreed, honestly I put it here just to remember me to do something with it later!

layout/layout_test.go
natalieparellano: I tried to move this to another file, but there is work to do to avoid sharing the docker registry

natalieparellano: Another comment, many of the tests are pulling directly from Docker Hub, which causes us to hit the rate limit fairly easily. Something to possibly refactor

layout/new_test.go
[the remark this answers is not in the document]
natalieparellano: Maybe resolved by https://github.com/husni-faiz/imgutil/pull/12

remote/new_test.go
natalieparellano: Maybe resolved by https://github.com/husni-faiz/imgutil/pull/12

remote/new.go
natalieparellano: Just a note, we're not actually validating the "repo name" for remote __images__ today (though maybe we should). I think I removed it in https://github.com/husni-faiz/imgutil/pull/12, but we could add it back.

layout/new.go
natalieparellano: I deleted this code block in https://github.com/husni-faiz/imgutil/pull/12, but I am not sure if I fixed

> If the index is not found, it does nothing and we are throwing an error

layout/new.go
natalieparellano: Partially resolved by https://github.com/husni-faiz/imgutil/pull/12 (changed the signature)

new.go
natalieparellano: Resolved by https://github.com/husni-faiz/imgutil/pull/12

layout/new.go
natalieparellano: I started to do this in https://github.com/husni-faiz/imgutil/pull/12 but then I saw some value in at least returning a CNB image from the constructor, since we are formatting the options to the extent that we set the base index "instance". Also have it work the same way as the image constructor is convenient

remote/new.go
natalieparellano: Same comment as https://github.com/buildpacks/imgutil/pull/191#discussion_r1575392890

cnb_image.go
jjbustamante: In the initial requirement we were expecting to set only `os`, `architecture`, `variant` information in a manifest that belongs to an Index. Sai exposed this attribute with the `pack manifest add` and `pack manifest annotate` commands, but I do not think the community is requesting to set these values. Yes, they are in the OCI spec, but so far, nobody is asking for it

jjbustamante: it seems this is useful for `windows` indexes, I think our plans for windows support are very limited

layout/layout_test.go
jjbustamante: I know it is confusing, but we wanted to allow to push an `layout` and `remote` index

cnb_index.go
```diff
@@ -35,161 +35,144 @@ type CNBIndex struct {
...
 
 // OS returns `OS` of an existing Image.
 func (h *CNBIndex) OS(digest name.Digest) (os string, err error) {
-	configFile, err := h.getConfigFileFrom(digest)
+	desc, err := h.getDescriptorFrom(digest)
 	if err != nil {
 		return "", err
 	}
-	return configFile.OS, nil
+	if desc.Platform != nil {
```
natalieparellano, line 57: nit: `getDescriptorFrom` never returns nil in its first argument, so I think you can remove this check

jjbustamante: I think I added it because on unit testing when using `random` packages in `ggcr` some nil in these attributes but let me check if everything is file after removing it

layout/layout_test.go
natalieparellano: I believe this was fixed by @jjbustamante

cnb_index.go
```diff
@@ -0,0 +1,394 @@
...
+
+	var taggableIndex = NewTaggableIndex(indexManifest)
+	multiWriteTagables := map[name.Reference]remote.Taggable{
+		ref: taggableIndex,
+	}
+	for _, tag := range pushOps.DestinationTags {
+		multiWriteTagables[ref.Context().Tag(tag)] = taggableIndex
+	}
+
+	// FIXME: this will only push the index manifest, assuming that all the images it refers to exist in the registry
```
natalieparellano, line 324:
```suggestion
	// Note: this will only push the index manifest, assuming that all the images it refers to exist in the registry
```

natalieparellano:
@jabrown85 I think this one is just about ready if you'd like to take a look. I turned on squashing for PRs so that we can bring this into main as one commit 😅

layout/layout_test.go
jjbustamante: Yes, I think I moved the test to `index_test.go`

jabrown85 (approved):

PR #193, opened 2023-04-11
fixes linter failing on TypeRegA

joe-kimmel-vmw:


local/save_file.go
```diff
@@ -184,7 +184,7 @@ func untar(r io.Reader, dest string) error {
 			if err := os.MkdirAll(path, hdr.FileInfo().Mode()); err != nil {
 				return err
 			}
-		case tar.TypeReg, tar.TypeRegA:
```
joe-kimmel-vmw, line 187: I'm not sure what's going on but the linter I have locally fails on `TypeRegA` so I can't run `make test` at all without this change...   seems safe since the failure is that `TypeRegA` has been deprecated for a long time and nobody should be using it.

natalieparellano (approved):

PR #202, opened 2023-05-15
Add the ability for images to retain history if provided

natalieparellano:
This is a first step toward https://github.com/buildpacks/lifecycle/issues/411

I think this does more or less what is expected, but I need to add tests 😈

### DIFF
--- a/image.go
+++ b/image.go
@@ -58,6 +58,7 @@ type Image interface {
 
 	AddLayer(path string) error
 	AddLayerWithDiffID(path, diffID string) error
+	AddLayerWithDiffIDAndHistory(path, diffID string, history v1.History) error
 	Delete() error
 	Rebase(string, Image) error
 	RemoveLabel(string) error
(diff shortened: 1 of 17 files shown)


remote/new.go
```diff
@@ -296,9 +297,19 @@ func processBaseImageOption(ri *Image, baseImageRepoName string, platform imguti
...
+	if err != nil {
+		return err
+	}
+	layers, err := baseImage.Layers()
+	if err != nil {
+		return err
+	}
+	if len(configFile.History) != len(layers) {
+		configFile.History = make([]v1.History, len(layers))
+	}
```
natalieparellano, line 310: TODO: do this for local and layout images too

remote/save.go
```diff
@@ -35,11 +35,11 @@ func (i *Image) SaveAs(name string, additionalNames ...string) error {
 	if err != nil {
 		return errors.Wrap(err, "get image layers")
 	}
-	cfg.History = make([]v1.History, len(layers))
+	if len(cfg.History) != len(layers) {
+		cfg.History = make([]v1.History, len(layers))
+	}
```
natalieparellano, line 40: We shouldn't need this as we've been careful to update the history when instantiating an image and adding layers, but you never know...

local/new.go
```diff
@@ -148,6 +150,7 @@ func processBaseImageOption(image *Image, baseImageRepoName string, platform img
 	}
 
 	image.inspect = inspect
+	image.history = make([]v1.History, len(image.inspect.RootFS.Layers))
```
natalieparellano, line 153: TODO: we should grab the actual history

local/new.go
```diff
@@ -45,6 +46,7 @@ func NewImage(repoName string, dockerClient DockerClient, ops ...ImageOption) (*
 		docker:           dockerClient,
 		repoName:         repoName,
 		inspect:          inspect,
+		history:          make([]v1.History, len(inspect.RootFS.Layers)),
```
natalieparellano, line 49: TODO: we should grab the actual history and only zero it out if the length of the history doesn't match the length of the layers

image.go
```diff
@@ -123,50 +126,79 @@ func (t MediaTypes) LayerType() types.MediaType {
...
+func OverrideHistoryIfNeeded(image v1.Image) (v1.Image, error) {
+	configFile, err := image.ConfigFile()
+	if err != nil {
+		return nil, err
+	}
+	layers, err := image.Layers()
+	if err != nil {
+		return nil, err
+	}
+	retImage := image
```
joe-kimmel-vmw, line 179: ok this is maybe my own confusion(s) here.  but `v1.Image` is (probably) an interface, right?  which means here you're doing a pointer copy, right?  so below when you call `mutate` on the `retImage` doesn't that _also_ change the `image` param that was passed in?   (that would be a reason to do a deepcopy here,  imo, if you're trying not to change your input param)

natalieparellano: Summarizing offline discussion - I think we agreed that this is weird, but it is the way to do it unless you want to use a pointer to interface, which is also weird

remote/new.go
```diff
@@ -154,7 +156,17 @@ func processPreviousImageOption(ri *Image, prevImageRepoName string, platform im
...
 
+	configFile, err := prevImage.ConfigFile()
+	if err != nil {
+		return err
+	}
+
 	ri.prevLayers = prevLayers
+	prevHistory := configFile.History
+	if len(prevLayers) != len(prevHistory) {
+		prevHistory = make([]v1.History, len(prevLayers))
```
natalieparellano, line 167: Alternatively we could see if removing history for "empty layers" will cause the lengths to be equal

image.go
```diff
@@ -123,50 +127,88 @@ func (t MediaTypes) LayerType() types.MediaType {
...
 }
 
 // layersAddendum creates an Addendum array with the given layers
 // and the desired media type
-func layersAddendum(layers []v1.Layer, mediaType types.MediaType) []mutate.Addendum {
+func layersAddendum(layers []v1.Layer, history []v1.History, mediaType types.MediaType) []mutate.Addendum {
 	additions := make([]mutate.Addendum, 0)
-	for _, layer := range layers {
+	if len(history) != len(layers) {
+		history = make([]v1.History, len(layers))
```
jabrown85, line 205: Will this lead to an off by N type of association if we get here with partial history? Not super worried about that but curious.

jabrown85 (approved):
This was more work than I originally expected. Great stuff!

image.go
natalieparellano: I think we should be okay... we "normalize" the history (remove empty layers, check that the lengths are equal) at every stage - on initialization, when adding a layer, on save.

PR #210, opened 2023-06-26
When checking read/write access, if there is no access, surface the e…

natalieparellano:
…rror that occurred

This can be helpful in debugging failed builds where access is expected

### DIFF
--- a/remote/remote_test.go
+++ b/remote/remote_test.go
@@ -1927,7 +1927,9 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 			it("returns true", func() {
 				image, err := remote.NewImage(repoName, authn.DefaultKeychain)
 				h.AssertNil(t, err)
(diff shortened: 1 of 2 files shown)


remote/remote_test.go
```diff
@@ -1968,31 +1973,37 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+					h.AssertEq(t, canRead, true)
 				})
 			})
 
 			when("image doesn't have read access but has write access", func() {
 				it("returns false", func() {
 					image, err := remote.NewImage(customRegistry.RepoName(writeOnlyImage), authn.DefaultKeychain)
 					h.AssertNil(t, err)
-					h.AssertEq(t, image.CheckReadAccess(), false)
+					canReadWrite, _ := image.CheckReadWriteAccess()
```
natalieparellano, line 1996:
```suggestion
					canReadWrite, _ := image.CheckReadAccess()
```

PR #214, opened 2023-07-20
Fixed version of the linter 1.51.1

dlion:
## Summary
I noticed that on [Lifecycle](https://github.com/buildpacks/lifecycle/blob/main/Makefile#L260) we are using a fixed version of the linter but on [Imgutil](https://github.com/buildpacks/imgutil/blob/main/Makefile#L18) we are using the latest one.
It can be risky having always the latest version (introduction of bugs and breaking changes)
So to have it aligned I added the fixed version on imgutil Makefile

#### Before

Whenever we download the linter we install the latest version

#### After

Now whenever we download the linter we install version 1.51.1 

## Documentation

    - [ ] Yes, see #___
    - [X] No

## Related

Resolves #___

### DIFF
--- a/fakes/image_test.go
+++ b/fakes/image_test.go
@@ -3,12 +3,14 @@ package fakes_test
 import (
 	"archive/tar"
 	"fmt"
-	"github.com/sclevine/spec"
-	"github.com/sclevine/spec/report"
+
 	"os"
 	"path/filepath"
 	"sort"
 	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
 
 	"github.com/buildpacks/imgutil"
 	"github.com/buildpacks/imgutil/fakes"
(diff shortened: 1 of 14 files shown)


natalieparellano:
@dlion thanks for the PR! I think we need to fix these lint errors now :)

fakes/image_test.go
```diff
@@ -3,16 +3,12 @@ package fakes_test
 import (
 	"archive/tar"
 	"fmt"
-	"io/ioutil"
-	"math/rand"
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
```
natalieparellano, line 7: nit: these should go back into their own group

natalieparellano (approved):
LGTM @dlion - I left one nit

fakes/image_test.go
dlion: Uhm, IntelliJ automatically formatted it in that way, I'll fix it right away

dlion:
Done :) @natalieparellano

PR #218, opened 2023-08-29
Set InsecureSkipVerify flag as the Insecure one

dlion:
This PR concerns the [insecure-registry RFC](https://github.com/buildpacks/rfcs/blob/main/text/0111-support-insecure-registries.md) implementation in the imgutil library.

In the [previous PR](https://github.com/buildpacks/imgutil/pull/154), a function`WithRegistrySetting` has been implemented with the possibility to activate or deactivate the `InsecureSkipVerify` flag.

[The RFC](https://github.com/buildpacks/rfcs/blob/main/text/0111-support-insecure-registries.md#how-it-works) wasn't very clear about the reason it needed to be permanently set as false/true so I tried to investigate a bit how other tools/libraries implemented it.
From Google:
```
// Insecure is an Option that allows image references to be fetched without TLS.
// This will also allow for untrusted (e.g. self-signed) certificates in cases where
// the default transport is used (i.e. when WithTransport is not used).
```
This implementation followed the one made by Google: https://github.com/google/go-containerregistry/blob/main/pkg/crane/options.go#L70 and https://github.com/google/go-containerregistry/blob/190ad0e4d556f199a07951d55124f8a394ebccd9/cmd/crane/cmd/root.go#L83 

With this PR, whenever we set a registry as `Insecure` we also disable the TLS verification.

### DIFF
--- a/acceptance/reproducibility_test.go
+++ b/acceptance/reproducibility_test.go
@@ -3,14 +3,15 @@ package acceptance
 import (
 	"context"
 	"fmt"
+	"os"
+	"testing"
+
 	dockerclient "github.com/docker/docker/client"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	ggcrremote "github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/sclevine/spec"
 	"github.com/sclevine/spec/report"
-	"os"
-	"testing"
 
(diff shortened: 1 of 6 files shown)


natalieparellano (approved):
Love it, thanks for taking the time to research how other tools are handling this @dlion!

@jabrown85 did you want to have a look also?

jabrown85 (approved):

PR #219, opened 2023-09-15
Add insecureSkipVerify in other missing functions

dlion:
@jabrown85 [has spotted](https://github.com/buildpacks/lifecycle/pull/1140#issuecomment-1719849393) some points in the `imgutil` library where the `insecureSkipVerify` flag should be necessary.

This PR is addressing that.

Thank you @jabrown85!

### DIFF
--- a/remote/new.go
+++ b/remote/new.go
@@ -1,7 +1,6 @@
 package remote
 
 import (
-	"crypto/tls"
 	"io"
 	"net/http"
 	"strings"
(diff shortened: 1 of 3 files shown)


jabrown85:
From the outside I would expect to be able to do this and have the `remote.WithTransport` use the correct `reg.insecure` value. 
```golang
	remoteImage, err := remote.NewImage("myregistry.domain.com/repo/app-image",
		authn.DefaultKeychain,
		remote.FromBaseImage("myother-insecure-registry.com/repo/superbase"),
		remote.WithRegistrySetting("myregistry.domain.com", true),
		remote.WithRegistrySetting("myother-insecure-registry.com", true),
	)
```

I see `map[string]registrySetting` already in `getRegistry` - If we update `remote.WithRegistrySetting` to not _override_ the entire map it _may_ just work. Everything reads ok to me.

jabrown85:
@natalieparellano I will leave for you to review as well.

dlion:
@jabrown85 now I got it! Thanks for clarifying.
Yes you are right, the actual implementation reset the map every time, preventing having multiple registries, and the test we currently have test just one registry. 😅  I'll update this PR fixing the map initialization and extending the test with the check for multiple registries. 

Thank you again for spotting it 😁

natalieparellano (approved):
Looks good to me! Thanks for the sharp eyes @jabrown85 :)

PR #221, opened 2023-10-09
`pack builder create` should accept builder env config

WYGIN:


natalieparellano:
@WYGIN I'm a bit confused, why is this needed? I see `SetBuildConfigEnv` called on `builder.Builder` in https://github.com/buildpacks/pack/pull/1926, but that's not an `imgutil.Image`.

PR #222, opened 2023-10-25
Fix local images when daemon uses containerd storage

natalieparellano:
In service of:
* https://github.com/buildpacks/pack/issues/1519
* https://github.com/buildpacks/lifecycle/issues/829

On Mac: Docker > Settings > Features in development > Use containerd for pulling and storing images
Tests pass for me locally with both configurations.

Local image creation works with these changes, but is unsatisfactory as we can no longer omit base layers that are already present in the daemon. Omitting the layers creates a runnable image, but if you try to `docker push` that image to a registry the daemon will crash, I'm not sure why. But if you do a `docker save` on that image you will get a base layer that is 0 in size. It's possible this is something that will be fixed before containerd storage becomes the default configuration. We should keep an eye on it (or maybe there's another way around it).


acceptance/reproducibility_test.go
```diff
@@ -164,6 +164,12 @@ func compare(t *testing.T, img1, img2 string) {
 	cfg2, err := v1img2.ConfigFile()
 	h.AssertNil(t, err)
 
+	// images that were created locally may have `DockerVersion` equal to "dev" and missing `Config.Image`
```
natalieparellano, line 167:
```suggestion
	// images that were created locally may have `DockerVersion` equal to "dev" and missing `Config.Image` if the daemon uses containerd storage
```

acceptance/reproducibility_test.go
```diff
@@ -164,6 +164,12 @@ func compare(t *testing.T, img1, img2 string) {
...
+	cfg1.DockerVersion = ""
+	cfg2.DockerVersion = ""
```
natalieparellano, line 169: I tried explicitly setting this about here: https://github.com/buildpacks/imgutil/blob/67824e989715b0bdd3793f4df3d2e55d643052d8/local/save_file.go#L210 but it seems to have no effect.

testhelpers/testhelpers.go
```diff
@@ -206,25 +206,28 @@ func DockerRmi(dockerCli dockercli.CommonAPIClient, repoNames ...string) error {
 
 // PushImage pushes an image to a registry, optionally using credentials from any set DOCKER_CONFIG
 func PushImage(t *testing.T, dockerCli dockercli.CommonAPIClient, refStr string) {
-	ref, err := name.ParseReference(refStr, name.WeakValidation)
-	AssertNil(t, err)
+	t.Helper()
+	Run(t, exec.Command("docker", "push", refStr)) // #nosec G204
```
natalieparellano, line 210: This was interesting. `dockerCli.ImagePush` returned a 401 error, but `docker image push` worked fine. I didn't look too deeply into it, but this is another rough edge that could be worth investigating.

local/save.go
```diff
@@ -154,13 +123,16 @@ func (i *Image) doSaveAs(name string) (types.ImageInspect, error) {
 		return types.ImageInspect{}, errors.Wrapf(err, "loading image %q. first error", i.repoName)
 	}
 
-	inspect, _, err := i.docker.ImageInspectWithRaw(context.Background(), id)
+	inspect, _, err := i.docker.ImageInspectWithRaw(context.Background(), i.repoName)
```
natalieparellano, line 126: With containerd storage the image ID is the sha of the manifest, which technically we could compute but it would be hard. To avoid collision problems we can validate that the data in inspect matches the config that we expect.

local/save.go
```diff
@@ -24,10 +25,17 @@ func (i *Image) Save(additionalNames ...string) error {
...
+		inspect types.ImageInspect
+		err     error
+	)
+	canOmitBaseLayers := !usesContainerdStorage(i.docker)
+	if canOmitBaseLayers {
+		// During the first save attempt some layers may be excluded.
+		// The docker daemon allows this if the given set of layers already exists in the daemon in the given order.
+		inspect, err = i.doSaveAs(name)
+	}
+	if !canOmitBaseLayers || err != nil {
```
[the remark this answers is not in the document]
jabrown85: I also couldn't find a way around this. The trick of giving a tar header with no bytes does not work so we always have to pay the price it seems. At least until https://github.com/moby/moby/issues/44369 ships

testhelpers/testhelpers.go
jabrown85: That is very weird. Maybe we should open a bug on this one. I captured the flow a bit.

```
// our docker.Push wrapper
	fmt.Println("DEBUG (DockerPush)", authConfig.Username, authConfig.Password)

	rc, err := dockerCli.ImagePush(context.Background(), refStr, dockertypes.ImagePushOptions{
		RegistryAuth: base64.URLEncoding.EncodeToString(encodedJSON),
		PrivilegeFunc: func() (string, error) {
			fmt.Println("DEBUG (DockerPush)", "PrivilegeFunc")
			return "bob", nil
		},
	})

// the registry bit
func BasicAuth(handler http.Handler, username, password, realm string) http.Handler {
		user, pass, ok := r.BasicAuth()
		fmt.Println("DEBUG (RegistryBasicAuth)", "user:", user, "pass:", pass, "ok:", ok)

// relevant debug output
DEBUG (DockerPush) vvqijndzyz wiqldsvicu
DEBUG (RegistryBasicAuth) user:  pass:  ok: false
DEBUG (RegistryBasicAuth) user:  pass:  ok: false
DEBUG (RegistryBasicAuth) user:  pass:  ok: false
```

The generated auth is passed into `ImagePushOptions. The first empty user/pass at the registry auth check is typical but the subsequent ones are not. I tried adding the `PrivilegeFunc` to see if the 401 would trigger an auth send but the logs show it isn't executed in either situation. Like the auth is ignored entirely 😞 

Here are the logs when using the overlayfs for comparison.

```
DEBUG (DockerPush) zitozndflk dhlzgeqnku
DEBUG (RegistryBasicAuth) user:  pass:  ok: false
DEBUG (RegistryBasicAuth) user: zitozndflk pass: dhlzgeqnku ok: true
DEBUG (RegistryBasicAuth) user: zitozndflk pass: dhlzgeqnku ok: true
```

jabrown85 (approved):
Great work! This seems shippable for me since the storage system is in beta on the docker side.

PR #226, opened 2023-10-27
Fix indentation in Dependabot config

edmorley:
The `groups` section added in #223 was indented incorrectly, resulting in Dependabot ignoring the `groups` section of the config.

It's not clear why the Dependabot config schema validation didn't highlight the error on the PR - perhaps because it doesn't run on forked PRs?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,12 +4,12 @@ updates:
   directory: "/"
   schedule:
     interval: weekly
-    groups:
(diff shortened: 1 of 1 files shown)


edmorley:
@natalieparellano Apologies for missing this in the original PR. I've checked all of my other unmerged Dependabot update PRs and they don't have the same issue.

natalieparellano (approved):
Thank you @edmorley!

PR #232, opened 2023-11-01
Use daemon or runtime provided arch instead of amd64

jericop:
The current `defaultPlatform` functions sets the architecture to "amd64", which causes pack to use the `amd64` build image, and to install the `amd64` lifecycle, when creating a new builder from an `arm64` host with no docker daemon installed. 

This changes the default platform functions to use the docker daemon os and architecture for local images, and `runtime.GOARCH` as the architecture for new remote images.

The following log file is the output of a script that was used to identify the issue.
https://github.com/jericop/github-actions/blob/main/buildpacks/create-multi-arch-builders/standalone/imgutil-pr-232-verify-amd64-issue.sh.log

You can see that the script doesn't finish. You can also see that it downloads an `amd64` lifecycle rather than the `arm64` as expected.
https://github.com/jericop/github-actions/blob/main/buildpacks/create-multi-arch-builders/standalone/imgutil-pr-232-verify-amd64-issue.sh.log#L159

The following log file is the output of a script was used to identify that an updated version of pack that uses this fix.
https://github.com/jericop/github-actions/blob/main/buildpacks/create-multi-arch-builders/standalone/imgutil-pr-232-verify-fix.sh.log

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -41,6 +41,7 @@ type DockerClient interface {
 	ImageSave(ctx context.Context, images []string) (io.ReadCloser, error)
 	ImageTag(ctx context.Context, image, ref string) error
 	Info(ctx context.Context) (types.Info, error)
+	ServerVersion(ctx context.Context) (types.Version, error)
 }
 
 // getters
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -43,6 +43,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 	var (
 		dockerClient          client.CommonAPIClient
 		daemonOS              string
+		daemonArchitecture    string
 		runnableBaseImageName string
 	)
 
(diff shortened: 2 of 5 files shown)


jjbustamante:
## Summary

This PR is fixing an issue when running pack using emulation with `docker buildx` which I think it is great!!, when running pack using emulation. 

Additionally, I think the root problem is that we can't define the platform when creating Builders, something we are trying to solve with the multi-arch [RFC](https://github.com/buildpacks/rfcs/blob/jjbustamante/feature/multi-arch-phase-2/text/0000-multiarch-builders-and-package.md) but improving the default values to be more realistic than hard coding something, I think is a great improvement

## Local change

-  We are changing from using the [/info](https://docs.docker.com/engine/api/v1.43/#tag/System/operation/SystemInfo) API, and reading `OSType`:  type of the operating system of the host, as returned by the Go runtime (GOOS). Currently returned values are "linux" and "windows", and hard coding `amd64`
- To use [/version](https://docs.docker.com/engine/api/v1.43/#tag/System/operation/SystemVersion) API, and reading 
   - `Os` The operating system that the daemon is running on ("linux" or "windows")
   - `Arch` The architecture that the daemon is running on
   This makes a lot of sense to me, because with the local interface we are going to look for images in the daemon. 

## Remote change
- We are changing the hard coded `amd64` for `runtime.GOARCH` which is defined as: is the running program's architecture target: one of 386, amd64, arm, s390x, and so on. 

  This also makes sense to me, because when we are running pack commands, we expect to match the host machine

jjbustamante (approved):
I tested Jerico script and check `pack` running and everything looks good to me!

@natalieparellano if you can take a look will be awesome!

natalieparellano (approved):

PR #233, opened 2023-11-02
sparse: do not modify media type when unspecified

jabrown85:
The media type should only be modified when the user specifies it. In my case I noticed the digest given by the sparse.Image was not the same as the image it was dervied from. This was because the logic to set the media type defaulted to overriding to OCI if the media type was not specified. This commit changes the logic to only override the media type if the user specifies it.

### DIFF
--- a/layout/layout_test.go
+++ b/layout/layout_test.go
@@ -157,7 +157,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 					it("sets the initial state from a linux/arm base image", func() {
 						existingLayerSha := "sha256:5a0b973aa300cd2650869fd76d8546b361fcd6dfc77bd37b9d4f082cca9874e4"
 
-						img, err := layout.NewImage(imagePath, layout.FromBaseImage(testImage))
+						img, err := layout.NewImage(imagePath, layout.FromBaseImage(testImage), layout.WithMediaTypes(imgutil.OCITypes))
 						h.AssertNil(t, err)
 						h.AssertOCIMediaTypes(t, img)
 
@@ -199,7 +199,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 					img, err := layout.NewImage(imagePath, layout.FromBaseImagePath(fullBaseImagePath))
 					h.AssertNil(t, err)
-					h.AssertOCIMediaTypes(t, img)
+					h.AssertDockerMediaTypes(t, img)
 
 					os, err := img.OS()
 					h.AssertNil(t, err)
@@ -225,7 +225,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 					img, err := layout.NewImage(imagePath, layout.FromBaseImagePath(sparseBaseImagePath))
 					h.AssertNil(t, err)
-					h.AssertOCIMediaTypes(t, img)
+					h.AssertDockerMediaTypes(t, img)
 
 					os, err := img.OS()
 					h.AssertNil(t, err)
--- a/layout/new.go
+++ b/layout/new.go
@@ -59,10 +59,10 @@ func NewImage(path string, ops ...ImageOption) (*Image, error) {
 		ri.createdAt = imageOpts.createdAt
 	}
 
-	if imageOpts.mediaTypes == imgutil.MissingTypes && !hasBaseImage {
-		ri.requestedMediaTypes = imgutil.OCITypes
-	} else {
+	if imageOpts.mediaTypes != imgutil.MissingTypes {
 		ri.requestedMediaTypes = imageOpts.mediaTypes
+	} else if !hasBaseImage {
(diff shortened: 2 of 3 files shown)


layout/layout_test.go
```diff
@@ -157,7 +157,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 					it("sets the initial state from a linux/arm base image", func() {
 						existingLayerSha := "sha256:5a0b973aa300cd2650869fd76d8546b361fcd6dfc77bd37b9d4f082cca9874e4"
 
-						img, err := layout.NewImage(imagePath, layout.FromBaseImage(testImage))
+						img, err := layout.NewImage(imagePath, layout.FromBaseImage(testImage), layout.WithMediaTypes(imgutil.OCITypes))
```
jjbustamante, line 160: I am a little confuse now. 

The test is using the following image: `arm64v8/busybox@sha256:50edf1d080946c6a76989d1c3b0e753b62f7d9b5f5e66e88bef23ebbd1e9709c` which is using `Docker Media-Types`

Adding `layout.WithMediaTypes(imgutil.OCITypes)` I will expect to *override* it and my image should have `OCI Media-Types`.  Should line 202 doesn't change, right?

jabrown85: I think you need to expand the diff. Line 162 did _not_ change because of what you said. Line 202 is the result of line 200, which did _not_ use the `WithMediaTypes` override. I sort of mixed and matched these existing tests with OCI/Docker. Let me know if that doesn't make sense!

layout/new.go
```diff
@@ -58,7 +59,7 @@ func NewImage(path string, ops ...ImageOption) (*Image, error) {
 		ri.createdAt = imageOpts.createdAt
 	}
 
-	if imageOpts.mediaTypes == imgutil.MissingTypes {
+	if imageOpts.mediaTypes == imgutil.MissingTypes && !hasBaseImage {
```
jjbustamante, line 62: Following my previous comment. 
The test is providing a base image and the option to override the media type, so that's why the it is never getting into this line and the media-type is not overrided

jjbustamante: Maybe something like this should do the trick: ?
```go
	if imageOpts.mediaTypes != imgutil.MissingTypes {
		ri.requestedMediaTypes = imageOpts.mediaTypes
	} 
```

jabrown85: Yeah. The idea is that if `layout.WithMediaTypes` is omitted, the `imageOpts.mediaTypes` will be `imgutil.MissingTypes`. In the case where there was a base image (always true in sparse, not always true in layout) - we should default to matching the base image.

layout/layout_test.go
jjbustamante: Ohhh . sorry, let me check

jjbustamante: Sorry about that, you are right!! I was checking a bad line.

jjbustamante (approved):
Great job @jabrown85! This looks good to me!

layout/sparse/sparse_test.go
```diff
@@ -107,5 +109,51 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
+			it("mutates the media type to the specified media type", func() {
+				image, err := sparse.NewImage(imagePath, testImage, layout.WithMediaTypes(imgutil.OCITypes))
+				h.AssertNil(t, err)
+
+				err = image.Save()
+				h.AssertNil(t, err)
+
+				actualMediaType, err := image.MediaType()
+				h.AssertNil(t, err)
+				h.AssertEq(t, actualMediaType, types.MediaType("application/vnd.oci.image.manifest.v1+json"))
```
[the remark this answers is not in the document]
jabrown85: Good shout!

natalieparellano (commented):
Are we going to need to change the lifecycle at all? Or should that be covered by "has base image"?

jabrown85:
I don't think lifecycle is aware of sparse vs layout at all today and this should change nothing there.

jjbustamante:
> Are we going to need to change the lifecycle at all? Or should that be covered by "has base image"?

As I mentioned on Slack: 
- Originally I was always overriding the media types to use OCI because tooling like [umoci](https://umo.ci), which I used to verify the image was working, didn't recognize docker media types. 
- I would like to add a flag similar to `skopeo copy` to allow users to override the media type if they want to:
```bash
-f, --format string                         MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifest type of source, with fallbacks)
```
   - In this case, I think we will need to make some changes in [lifecycle](https://github.com/buildpacks/lifecycle/blob/33c5b890044980b25810c0719801af0398fe9b3b/cmd/lifecycle/exporter.go#L394-L439) and pack to pass through the media-type option, but I think this can just be an improvement.

natalieparellano (approved):

Issue #207, opened 2023-05-30
imgutil images should be v1.Images


natalieparellano:
Currently, a `layout.Image` implements the `v1.Image` interface (because the underlying "base" image is embedded). With some small changes, `remote.Image` could follow this pattern. `local.Image` is trickier because of the daemon, but with some work we could assemble enough information to return a `v1.ConfigFile`, erroring if a manifest is requested. We'll need to be careful to distinguish what is the "saved" image vs the "working image" (e.g., after `AddLayer` is called the layer is added to the working image but not the image that's saved in the disk/registry/daemon) but if we do it right we can remove a bunch of duplication across the layout/remote/local packages and make the interface more intuitive to work with.

sagnik3788:
@natalieparellano may I give it a try? It would be helpful for me if you could share some insights about this issue. Thank you

natalieparellano:
@sagnik3788 aye, I actually took this up before the winter break and it turned out to be VERY involved. I will probably put a PR up for it soon (I should probably have assigned myself the issue before) - when that is up I would appreciate any help in reviewing it, which would also be a great help :)

PR #238, opened 2024-01-10
Refactor: local images should be v1.Images

natalieparellano:
Partially resolves https://github.com/buildpacks/imgutil/issues/207

Introduces a `locallayout` package and a new `imgutil.CNBImage` struct. 

The `CNBImage` struct wraps a `v1.Image`, which could be `locallayout`, `remote`, `layout`, etc. and performs all the modifications to the image that are necessary for a CNB build (such as re-using a layer from a previous image).

`locallayout.Image` implements a `v1.Image` by wrapping output from the daemon. However, it does not contain any CNB logic. It does have code that is specific to constructing new images and saving them to the daemon. 

This should eventually allow us to delete duplicated code in the `remote` & `layout` packages that modifies images in those packages, as we can remove all of the CNB logic (and then some).

Additionally, this should make it easier for us to send images to the daemon in OCI layout format.

This PR simply adds code but if `locallayout.Image` works well in place of `local.Image` we can remove `local`.


locallayout/new.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/image"
+	"github.com/docker/docker/client"
+
+	"github.com/buildpacks/imgutil"
+)
+
+// NewImage returns a new image that can be modified and saved to a docker daemon
+// via a tarball in OCI layout format.
```
natalieparellano, line 16: This isn't true yet

natalieparellano:
@jabrown85 I would love to get your thoughts on this one 🙏🏼

jabrown85 (approved):
This sounds great to me!

PR #242, opened 2024-01-23
DRAFT: Send oci-layout formatted tars to the daemon

natalieparellano:


locallayout/image_test.go
```diff
@@ -1137,20 +1137,20 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 
 			it.Before(func() {
 				// new base image
-				newBase = "pack-newbase-test-" + h.RandString(10)
+				newBase = newTestImageName()
+				// TODO: this test consistently fails when the below line is commented in
```
natalieparellano, line 1141: I really don't know what to say about this

locallayout/store.go
```diff
@@ -165,55 +167,70 @@ func (s *Store) doSave(image v1.Image, withName string) (types.ImageInspect, err
...
 
 func (s *Store) addImageToTar(tw *tar.Writer, image v1.Image, withName string) error {
-	rawConfigFile, err := image.RawConfigFile()
+	path, err := os.MkdirTemp("", "") // TODO: remove?
 	if err != nil {
 		return err
 	}
-	configHash := fmt.Sprintf("%x", sha256.Sum256(rawConfigFile))
-	if err = addTextToTar(tw, rawConfigFile, configHash+".json"); err != nil {
+	layoutPath, err := layout.Write(path, empty.Index)
```
natalieparellano, line 174: It would be more efficient to add content to the tar as it's being streamed, but that requires more work

locallayout/store.go
```diff
@@ -165,55 +167,70 @@ func (s *Store) doSave(image v1.Image, withName string) (types.ImageInspect, err
...
+	if err != nil {
 		return err
 	}
-	layers, err := image.Layers()
-	if err != nil {
+	if err = layoutPath.AppendImage(image, layout.WithAnnotations(map[string]string{"io.containerd.image.name": withName})); err != nil {
```
natalieparellano, line 178: It would be more efficient to add content to the tar as it's being streamed, but that requires more work

locallayout/v1_facade.go
```diff
@@ -197,8 +209,8 @@ func toV1Config(dockerCfg *container.Config) v1.Config {
 var _ v1.Layer = &v1LayerFacade{}
 
 type v1LayerFacade struct {
-	diffID            v1.Hash
-	optionalLayerPath string
+	diffID                  v1.Hash
+	optionalUnderlyingLayer v1.Layer
```
natalieparellano, line 213: This allows us to return the correct digest and compressed bits

cnb_image.go
```diff
@@ -203,19 +203,28 @@ func (i *CNBImageCore) WorkingDir() (string, error) {
...
 
+// TODO: this annotates the manifest, but maybe we need to annotate the __index__
 func (i *CNBImageCore) AnnotateRefName(refName string) error {
-	manifest, err := getManifest(i.Image)
+	var err error
+	i.Image, err = Annotate(i.Image, "org.opencontainers.image.ref.name", refName)
+	return err
+}
+
+func Annotate(image v1.Image, key, val string) (v1.Image, error) {
```
natalieparellano, line 213: Unrelated refactor

locallayout/v1_facade.go
```diff
@@ -212,43 +224,41 @@ func newEmptyLayerListFrom(configFile *v1.ConfigFile) []v1.Layer {
 }
 
 func (l v1LayerFacade) Digest() (v1.Hash, error) {
-	return l.diffID, nil
+	if l.optionalUnderlyingLayer != nil {
+		return l.optionalUnderlyingLayer.Digest()
+	}
+	return v1.NewHash("sha256:90e01955edcd85dac7985b72a8374545eac617ccdddcc992b732e43cd42534af")
```
natalieparellano, line 230: This and `Size()` below is the big bummer - we can't construct a valid manifest without it. But, we can't know it without pulling all the layers from the daemon. Many sighs

natalieparellano:
I'm going to close this PR for now. Until we have some way of getting the digest & compressed size of layers that already exist in the daemon (see https://github.com/moby/moby/issues/44369), we won't be able to construct a valid manifest for the image we're trying to load. However if we do get this information, we should be able to send an OCI layout tar that simply omits existing layer data in the blobs directory to avoid re-uploading the layers.

PR #243, opened 2024-01-26
Layout wraps v1 image

natalieparellano:


cnb_image.go
```diff
@@ -346,9 +333,6 @@ func (i *CNBImageCore) AddLayerWithDiffIDAndHistory(path, _ string, history v1.H
 }
 
 func (i *CNBImageCore) Rebase(baseTopLayerDiffID string, withNewBase Image) error {
-	if i.Kind() != withNewBase.Kind() {
```
jjbustamante, line 349: If my underlying image i is a remote image, does it mean I can rebase it with a local or layout image?

layout/options.go
```diff
@@ -8,79 +8,58 @@ import (
...
-	prevImagePath string
-	withHistory   bool
-	createdAt     time.Time
-	mediaTypes    imgutil.MediaTypes
+func FromBaseImagePath(name string) func(*imgutil.ImageOptions) {
+	return func(o *imgutil.ImageOptions) {
+		o.BaseImageRepoName = name
+	}
+}
+
```
jjbustamante, line 18: We used to have some documentation for each option, I think we should keep that.

jjbustamante:
I think you are still working in some **TODOs** but overall it looks good to me! As always, terrific job @natalieparellano!

cnb_image.go
```diff
@@ -18,36 +19,20 @@ import (
...
 
-type ImageStore interface {
-	Contains(identifier string) bool
-	Delete(identifier string) error
-	Save(image IdentifiableV1Image, withName string, withAdditionalNames ...string) (string, error)
-	SaveFile(image IdentifiableV1Image, withName string) (string, error)
-
-	DownloadLayersFor(identifier string) error
-	Layers() []v1.Layer
-}
```
natalieparellano, line 39: Originally I thought this could be an interface that all packages could implement, but since most packages would override the methods that call this interface it's not so useful

layout/save.go
```diff
@@ -17,69 +12,41 @@ func (i *Image) Save(additionalNames ...string) error {
...
 // SaveAs ignores the image `Name()` method and saves the image according to name & additional names provided to this method
 func (i *Image) SaveAs(name string, additionalNames ...string) error {
-	err := i.mutateCreatedAt(i.Image, v1.Time{Time: i.createdAt})
-	if err != nil {
-		return errors.Wrap(err, "set creation time")
-	}
-
-	if i.Image, err = imgutil.OverrideHistoryIfNeeded(i.Image); err != nil {
-		return fmt.Errorf("override history: %w", err)
+	if err := i.SetCreatedAtAndHistory(); err != nil {
```
natalieparellano, line 15: We have to do this at the point of save (vs instantiation) because otherwise we might modify the digest when there is a base image and that could be surprising

layout/sparse/sparse_test.go
```diff
@@ -141,9 +141,6 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 				image, err := sparse.NewImage(imagePath, testImage)
 				h.AssertNil(t, err)
 
-				err = image.Save()
-				h.AssertNil(t, err)
```
natalieparellano, line 145: Saving could mutate the digest and that is expected

new.go
```diff
@@ -6,62 +6,40 @@ import (
...
-
-	if !options.PreserveHistory {
-		if err = image.MutateConfigFile(func(c *v1.ConfigFile) {
-			for j := range c.History {
-				c.History[j] = v1.History{Created: v1.Time{Time: createdAt}}
-			}
-		}); err != nil {
-			return nil, err
-		}
-	}
```
natalieparellano, line 63: Moved this out of new method to avoid mutating the digest when there is a base image

natalieparellano (commented):
Worth noting, I tried importing this branch in pack and the lifecycle, and all tests are happy over there :)

jabrown85 (approved):

PR #249, opened 2024-02-06
Force the digest to be the same when saving sparse images

natalieparellano:
A sparse image is always constructed "from" another image, and this package is typically used to save "base" images (build or run) from the registry or wherever to disk.
In such scenarios, it is surprising for the digest to mutate when saving. In a sense, sparse images do not really need to be imgutil.Images at all, they just need to be v1.Images.

However for other local/remote/layout packages,
we typically want to zero out timestamps
in order to fulfill the reproducibility expectations of CNB, so preserving the digest when saving is not the default behavior.

### DIFF
--- a/layout/save.go
+++ b/layout/save.go
@@ -12,9 +12,12 @@ func (i *Image) Save(additionalNames ...string) error {
 
 // SaveAs ignores the image `Name()` method and saves the image according to name & additional names provided to this method
 func (i *Image) SaveAs(name string, additionalNames ...string) error {
-	if err := i.SetCreatedAtAndHistory(); err != nil {
-		return err
+	if !i.preserveDigest {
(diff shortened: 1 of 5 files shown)


layout/save.go
```diff
@@ -12,9 +12,12 @@ func (i *Image) Save(additionalNames ...string) error {
 
 // SaveAs ignores the image `Name()` method and saves the image according to name & additional names provided to this method
 func (i *Image) SaveAs(name string, additionalNames ...string) error {
-	if err := i.SetCreatedAtAndHistory(); err != nil {
-		return err
+	if !i.preserveDigest {
```
jjbustamante, line 15: Nice

jjbustamante (approved):

PR #251, opened 2024-03-01
Use the Image Digest to save it in a Registry

jjbustamante:
This change was done to replace the *image name* with the *digest* when saving it in a registry, the idea is to use it when we are creating multi-arch images and we need to save each intermediate image before creating the `Image Index`. 

The work is based on Husni's [PR](https://github.com/buildpacks/imgutil/pull/191), because I need his logic to create the `Image Index` but my change is actually only in this [commit](https://github.com/buildpacks/imgutil/commit/64e6affde1f4bb7902afea9bc196032dbc966a30)

This is a DRAFT! I am not sure if this is the best way to do it, I just needed a way to solve my problem and develop the PoC


jjbustamante:
We don't need this change anymore, I updated our pack manifest commands PR to use the latest version of the imgutil code (after we merged the image index logic) and we don't need this hack.

PR #256, opened 2024-04-01
Fix locallayout package, make it suitable to replace the local package

natalieparellano:
https://github.com/buildpacks/imgutil/pull/238 introduced a locallayout package with the idea that we would eventually use this package to send an OCI layout-formatted tar to the daemon instead of a legacy tar.

However, as described in https://github.com/buildpacks/imgutil/pull/242#issuecomment-1927692207 this isn't a viable approach for the time being.

This PR updates the locallayout package to remove the noted inefficiency and add support for containerd storage, so that locallayout can replace the local package. This will allow us to simplify our code because local images will be v1 Images.

### DIFF
--- a/locallayout/new.go
+++ b/locallayout/new.go
@@ -3,7 +3,6 @@ package locallayout
 import (
 	"context"
 	"fmt"
-	"sync"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/image"
@@ -48,7 +47,7 @@ func NewImage(repoName string, dockerClient DockerClient, ops ...func(*imgutil.I
 		baseIdentifier = processBase.identifier
 		store = processBase.layerStore
 	} else {
-		store = &Store{dockerClient: dockerClient, downloadOnce: &sync.Once{}}
+		store = NewStore(dockerClient)
 	}
 
 	cnbImage, err := imgutil.NewCNBImage(*options)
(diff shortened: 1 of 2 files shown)


natalieparellano:
This PR currently only speeds up getting the layer size when we download all the layers from the daemon - we need to record the size when layers are added

local/local_test.go
```diff
@@ -396,7 +395,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 		})
 
 		when("#WithConfig", func() {
-			var config = &container.Config{Entrypoint: []string{"some-entrypoint"}}
+			var config = &v1.Config{Entrypoint: []string{"some-entrypoint"}}
```
natalieparellano, line 398: This lifecycle (which uses `WithConfig`) will need to be updated before we can update imgutil

natalieparellano:
@jabrown85 @jjbustamante PTAL - this should reduce our maintenance overhead

jjbustamante (approved):
I think this is awesome!!! Thanks @natalieparellano for your hard work on simplifying the library.

local/local_test.go
```diff
@@ -1591,16 +1598,16 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
...
-				h.AssertError(t, err, `image "not-exist" does not contain layer with diff ID "some-layer"`)
+				h.AssertError(t, err, fmt.Sprintf("image %q does not contain layer with diff ID %q", "not-exist", someSHA))
 			})
 		})
 	})
 
 	when("#ReuseLayer", func() {
 		var (
-			prevImage     *local.Image
+			prevImage     imgutil.Image
```
natalieparellano, line 1610: I think we can change the constructor signature back to returning a pointer, so that this can still be `*local.Image`

natalieparellano:
I did some performance testing (using large images & layers) comparing the local package in this branch with the local package on main, and they seem equivalent. This should be good to go :)

jabrown85 (approved):

PR #257, opened 2024-04-02
Remote wraps v1

natalieparellano:
Branches off https://github.com/buildpacks/imgutil/pull/256


jabrown85 (approved):
Seems good to me as long as tests pass 😄

PR #266, opened 2024-04-12
Add interface method, AddOrReuseLayerWithHistory

natalieparellano:
The lifecycle currently does some bookkeeping to map a layer's "id" ("buildpack-id:layer-name") to the SHA of the layer in the previous image. It then compares this SHA to the SHA of the layer directory in the build container to determine if a layer should be added or re-used.

This works fine for layers that have an id (i.e., were created by a buildpack) but for extension layers we don't have a way to construct the mapping and simply want to add the layer if it's not present in the previous image, otherwise re-use it.

An imgutil.Image already has all the info necessary to determine whether a layer should be added or re-used, so adding an interface method is helpful for the lifecycle to avoid having to know anything about the previous image (at least for extension layers).

### DIFF
--- a/cnb_image.go
+++ b/cnb_image.go
@@ -357,6 +357,29 @@ func (i *CNBImageCore) AddLayerWithHistory(layer v1.Layer, history v1.History) e
 	return err
 }
 
+func (i *CNBImageCore) AddOrReuseLayerWithHistory(path string, diffID string, history v1.History) error {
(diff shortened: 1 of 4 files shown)


natalieparellano:
Branches off https://github.com/buildpacks/imgutil/pull/265

jabrown85 (approved):

PR #267, opened 2024-04-15
Fixes for imgutil from recent refactor

natalieparellano:
As surfaced in https://github.com/buildpacks/lifecycle/pull/1335

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -46,6 +46,11 @@ func (i *Image) Identifier() (imgutil.Identifier, error) {
 	}, nil
 }
 
+func (i *Image) Valid() bool {
+	// local images are actually always invalid, because they are missing a manifest, but let's ignore that for now
+	return true
+}
+
 // GetLayer returns an io.ReadCloser with uncompressed layer data.
 // The layer will always have data, even if that means downloading ALL the image layers from the daemon.
 func (i *Image) GetLayer(diffID string) (io.ReadCloser, error) {
--- a/cnb_image.go
+++ b/cnb_image.go
@@ -10,7 +10,6 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
-	"github.com/google/go-containerregistry/pkg/v1/validate"
 )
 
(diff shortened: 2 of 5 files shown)


local/local.go
```diff
@@ -46,6 +46,11 @@ func (i *Image) Identifier() (imgutil.Identifier, error) {
 	}, nil
 }
 
+func (i *Image) Valid() bool {
+	// local images are actually always invalid, because they are missing a manifest, but let's ignore that for now
```
natalieparellano, line 50: This was causing the analyzer not to ever find the previous image

local/store.go
```diff
@@ -216,24 +216,10 @@ func (s *Store) addImageToTar(tw *tar.Writer, image v1.Image, withName string) e
...
 		if err != nil {
 			return err
 		}
-		if size == -1 { // layer facade fronting empty layer
-			layerName = fmt.Sprintf("blank_%d", blankIdx)
-			blankIdx++
-			hdr := &tar.Header{Name: layerName, Mode: 0644, Size: 0}
-			if err := tw.WriteHeader(hdr); err != nil {
-				return err
-			}
```
natalieparellano, line 230: Moved to `s.addLayerToTar`

local/store.go
```diff
@@ -250,32 +236,46 @@ func (s *Store) addImageToTar(tw *tar.Writer, image v1.Image, withName string) e
 	return addTextToTar(tw, manifestJSON, "manifest.json")
 }
 
-func (s *Store) addLayerToTar(tw *tar.Writer, layer v1.Layer) (string, error) {
-	layerDiffID, err := layer.DiffID()
+func (s *Store) addLayerToTar(tw *tar.Writer, layer v1.Layer, blankIdx *int) (string, error) {
+	// If the layer is a previous image layer that hasn't been downloaded yet,
+	// cause ALL the previous image layers to be downloaded by grabbing the ReadCloser.
+	layerReader, err := layer.Uncompressed()
```
natalieparellano, line 242: Previously, we forced the layer to be downloaded by asking for its size.
    But, this has performance implications when you only need the size to construct a manifest.
    Now, we force the layer to be downloaded by opening a ReadCloser to the layer,
    and size only reports if the layer is present on disk, it doesn't cause the layer to be downloaded.

natalieparellano:
@jabrown85 @jjbustamante one more... 🤦🏼‍♀️

jabrown85 (approved):

PR #269, opened 2024-04-22
Ensure we always normalize history before saving

natalieparellano:
Previously, we normalized history whenever layers were added (either via AddLayer or ReuseLayer),
but when rebasing, if the new base image has empty layer history the information about the empty/non-empty aspect of the history will be lost if "preserve history" isn't set on the image.


jjbustamante (approved):
Great! This fixes a problem with were having on pack with the acceptance tests when [bumping](https://github.com/buildpacks/pack/pull/2136) to the latest version of lifecycle and imgutil

jabrown85 (approved):

PR #275, opened 2024-05-31
Fix: `archive/tar: write too long` error when using Docker Desktop containerd backend

natalieparellano:
Fixes (in part) https://github.com/buildpacks/pack/issues/2174

This is somewhat of a band-aid solution and degrades performance because we have to read the uncompressed layer bits to get the uncompressed layer size for layers that we `docker save` out of the daemon (re-used layers and all the base layers).

The better longer-term fix will be to send OCI-layout formatted tars when we have all the layers available (when using containerd as the backing store, this will always be true).

### DIFF
--- a/local/local.go
+++ b/local/local.go
@@ -1,13 +1,9 @@
 package local
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
-	"os"
-	"path/filepath"
 	"strings"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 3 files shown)


local/local.go
```diff
@@ -110,67 +106,29 @@ func (i *Image) SetOS(osVal string) error {
...
-	}
-	layer, err := i.addLayerToStore(path, diffID)
+	layer, err := i.store.AddLayer(path)
 	if err != nil {
 		return err
 	}
 	return i.AddLayerWithHistory(layer, emptyHistory)
 }
 
-func calculateChecksum(path string) (string, error) {
```
natalieparellano, line 124: This function relies on the daemon always outputting uncompressed layers, which is no longer true

local/store.go
```diff
@@ -546,9 +542,39 @@ func (s *Store) findLayer(withHash v1.Hash) v1.Layer {
...
+	if err != nil {
+		return nil, err
+	}
+	compressedSize, err := layer.Size()
+	if err != nil {
+		return nil, err
+	}
+	if fileSize == compressedSize {
+		// the layer is compressed, we don't know the uncompressed size
+		uncompressedSize = -1
```
natalieparellano, line 571: This is the real fix

local/store.go
```diff
@@ -288,11 +289,13 @@ func (s *Store) getLayerSize(layer v1.Layer) (int64, error) {
 		return 0, err
 	}
 	knownLayer, layerFound := s.onDiskLayersByDiffID[diffID]
-	if layerFound {
+	if layerFound && knownLayer.uncompressedSize != -1 {
```
natalieparellano, line 292: This is the real fix

jjbustamante (approved):
Can we add a unit test for this? or do we need docker 25 in the runner to be able to test it?

natalieparellano:
> do we need docker 25 in the runner to be able to test it?

Yeah, we need it. And, containerd should be enabled as the backing storage. It's a bit of a mess.

jabrown85 (approved):

PR #277, opened 2024-06-04
Fix podman disk issue: use proper index for blank layers

natalieparellano:
Before, we were not dereferencing the index pointer when constructing the layer name, resulting in layers named e.g., "blank_824644714416" when they should be "blank_0", "blank_1", and so on.

This is unfortunately very difficult to test without mocking out the docker client (so that we can record the stream that gets sent).


jabrown85 (approved):

natalieparellano:
Signed the merge commit, so that I can say that I "Manually verified the DCO"

PR #283, opened 2024-11-07
arm64 tests

joeybrown-sf:
Updated tests to make them runnable on arm64 machine.


local/local_test.go
```diff
@@ -717,7 +717,7 @@ func testImage(t *testing.T, when spec.G, it spec.S) {
 			h.AssertNil(t, err)
 
 			// based on static base image refs
-			expectedTime := time.Date(2018, 10, 2, 17, 19, 34, 239926273, time.UTC)
+			expectedTime := time.Date(2024, 9, 26, 21, 31, 42, 0, time.UTC)
```
natalieparellano, line 720: Just curious, this changed for linux/amd64 also, no? How does it still work if the linux/amd64 image digest stayed the same?

Issue #282, opened 2024-11-06
Add a warning message when saving a local image (daemon) and containerd storage is enable


jjbustamante:
There is a known performance issue when using containerd storage, #276 based on this [ticket](https://github.com/buildpacks/pack/issues/2272) on `pack` we want to add some warning message when saving an image to the daemon and using containerd

Some hints

- Add a new `WithLogger()` option [here](https://github.com/buildpacks/imgutil/blob/main/local/options.go)
- Modify the `Store` constructor, to also receive a `Logger` 
- Update the logic for `local.NewImage()` to check if we received the `Logger` option, in such as case, update the creation of the `Store` [here](https://github.com/buildpacks/imgutil/blob/main/local/new.go#L50) to pass through that Logger instance. 
- At this point, the `Store` has a valid logger instance to print the *warning* message during the [Image.Save ](https://github.com/buildpacks/imgutil/blob/main/local/store.go#L98-L108) operation without having to change the signature. It will depend on imageutil consumer to set the logger and if the logger is not set, we skip trying to print the *warning*, this must avoid breaking things.

edmorley:
A warning has since been added to Pack CLI instead:
https://github.com/buildpacks/pack/pull/2284

...so I think this issue can now be closed? (I don't have permissions to close it)

jjbustamante:
@edmorley I think you are right!

PR #285, opened 2024-11-11
Fix: imgutil/local: Add a warning message when saving a local image and containerd storage is enable #284

wuhenggongzi:
Fixse (imgutil) https://github.com/buildpacks/imgutil/issues/282

This pr enhances the Logger functionality by integrating it with the existing
automatic detection of the storage driver type in the NewStore function.
The Logger interface has been added to log warnings related to the containerd storage driver.
Test cases have been created to ensure that warnings are logged correctly when using containerd.

### DIFF
--- a/local/local_test.go
+++ b/local/local_test.go
@@ -27,6 +27,14 @@ const someSHA = "sha256:aec070645fe53ee3b3763059376134f058cc337247c978add178b6cc
 
 var localTestRegistry *h.DockerRegistry
 
+type TestLogger struct {
+	WarnCalled bool
+}
+
+func (logger *TestLogger) Warn(msg string) {
+	logger.WarnCalled = true
+}
+
 func TestLocal(t *testing.T) {
 	localTestRegistry = h.NewDockerRegistry()
 	localTestRegistry.Start(t)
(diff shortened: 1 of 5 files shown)


jjbustamante (commented):
@wuhenggongzi Thanks a lot for this work! once I saw it in code I had some thoughts about it,  but maybe @natalieparellano or @jabrown85 can take another look.

Wait a little to get more feedback before doing any change!

natalieparellano:
I'm a little wary that we're making this more complicated than it needs to be - I was able to add the warning with just a few lines of code here: https://github.com/buildpacks/pack/pull/2284/files

I guess, the downside is that when we fix this in imgutil (presumably when https://github.com/buildpacks/imgutil/pull/242#issuecomment-1927692207 is available via upstream docker) then we'll have to remember to remove the warning. But this could perhaps be made easier by creating various issues to capture the work mentioned in https://github.com/buildpacks/pack/issues/2272#issuecomment-2426435484 and then adding a note to the relevant issue in imgutil that we should remove the warning when fixed imgutil is part of pack.

edmorley:
A warning has since been added to Pack CLI instead:
https://github.com/buildpacks/pack/pull/2284

...so I think this PR can now be closed? (I don't have permissions to close it)

jjbustamante:
@wuhenggongzi Thanks a lot for this work! I am going to close it because we added the warning on `pack`

PR #287, opened 2025-03-31
chore: Bump github.com/docker/docker to v28.0.4 and fix broken API

alphanota:
Skaffold (github.com/googlecontainertools/skaffold) depends on `imgutil` as a dependency. We are trying to update to the latest docker package versions but unfortunately there are some braking changes between v26 and v28. This is my attempt at upgrading docker in imgutil and fixing the broken API.

### DIFF
--- a/acceptance/reproducibility_test.go
+++ b/acceptance/reproducibility_test.go
@@ -48,7 +48,7 @@ func testReproducibility(t *testing.T, _ spec.G, it spec.S) {
 		imageName1, imageName2 string
 		layer1, layer2         string
 		mutateAndSave          func(t *testing.T, image imgutil.Image)
-		dockerClient           dockerclient.CommonAPIClient
+		dockerClient           dockerclient.APIClient
 		runnableBaseImageName  string
 	)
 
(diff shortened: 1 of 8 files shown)


natalieparellano:
@alphanota thank you for this. We'll need to tread carefully here, as the CNB lifecycle also depends on imgutil, and the last time I tried to bump docker [there](https://github.com/buildpacks/lifecycle/pull/1414#issuecomment-2442500563) I ran into compatibility issues with buildkit and kaniko. Unfortunately bumping the buildkit version in kaniko will not be easy as kaniko has fallen into an [unmaintained](https://github.com/GoogleContainerTools/kaniko/issues/3348) state and that is a whole other story.

We are going to need to do something here, as we can't stay on an old docker version forever. I'll bring this up for discussion at one of our next Working Group meetings.

alphanota:
@natalieparellano Thanks @natalieparellano I closed this PR in favor of https://github.com/buildpacks/imgutil/pull/288 since I was having issues identifying all the lint warnings I needed to address in this PR.

PR #289, opened 2025-04-23
Fix: use proper transport for CheckReadWriteAccess

matejvasek:


matejvasek:
No idea why coverage is failing.

jjbustamante (approved):
Just for help on reviewing this change. 

```go
func GetTransport(insecure bool) http.RoundTripper {
	if insecure {
		return &http.Transport{
			TLSClientConfig: &tls.Config{
				InsecureSkipVerify: true, // #nosec G402
			},
		}
	}
	return http.DefaultTransport
}
```
If insecure is false, it is exactly like it is today

PR #291, opened 2025-06-26
Update docker and go-containerregistry

jericop:
This change does the following:
* Updates docker and go-containerregistry
* Changes toV1Config function input type to dockerspec.DockerOCIImageConfig because of a change in the latest version of docker/moby
* These fields are no longer supported by dockerCfg and are therefore no longer being set:
  - AttachStderr
  - AttachStdin
  - AttachStdout
  - Domainname
  - Hostname
  - Image
  - NetworkDisabled
  - OpenStdin
  - StdinOnce
  - Tty
  - ArgsEscaped is deprecated and has been removed as well

A corresponding PR to pack has been created, but it will need to be updated to point to the latest commit of imgutil once this change has been merged.
https://github.com/buildpacks/pack/pull/2406


go.mod
```diff
@@ -1,50 +1,53 @@
 module github.com/buildpacks/imgutil
 
 require (
-	github.com/docker/docker v28.1.1+incompatible
+	github.com/containerd/errdefs v1.0.0
+	github.com/docker/docker v28.2.2+incompatible
```
jabrown85, line 5: We may not be able to do this - we rely on kaniko and [it only goes to 28.1.1](https://github.com/GoogleContainerTools/kaniko/blob/main/go.mod#L14)

https://github.com/buildpacks/imgutil/pull/288#issuecomment-2770819530 for previous thoughts.

If lifecycle compiles with this commit, I'm happy to give it a go. I think before there were breaking changes but if the versions are compatible enough it could be safe.

jabrown85: @jericop I updated lifecycle to use the new chainguard fork of kaniko. This resolves the issue I brought up earlier. Get this PR passing checks and DCO. I re-kicked the github test action in case it was a fluke.

jericop: Nice! I figured out the issue and should have this passing now.

jabrown85 (approved):
This seems good to me

PR #294, opened 2025-08-13
Always try the fast path first (omitting base layers) regardless of storage type

jabrown85:
In my testing - trying to fast path works wonders on containerd storage layer now. I'm not entirely sure if this is because of upstream changes or other imgutil changes over time. An ephemeral builder creation goes from 6s back down to 10ms or less.

This is a partial revert of https://github.com/buildpacks/imgutil/pull/256

This fixes the first part of https://github.com/buildpacks/pack/issues/2272 for me. An untrusted ephemeral builder is much faster. The `image.Save` falls into the optimized path and doesn't have to do the slow `ensureLayers`

### DIFF
--- a/local/store.go
+++ b/local/store.go
@@ -260,7 +260,7 @@ func (s *Store) addImageToTar(tw *tar.Writer, image v1.Image, withName string, i
 		layerPaths []string
 		blankIdx   int
 	)
-	for i, layer := range layers {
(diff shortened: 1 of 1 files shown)


jkutner (approved):

edmorley:
Note: This PR was reverted in #297.

PR #305, opened 2026-02-17
Take two: Make local/docker implementation platform-aware

bojidar-bg:
Second attempt at #302 after https://github.com/buildpacks/lifecycle/issues/1605. 

Poking around the `failed to export: saving image: failed to fetch base layers:` error, I discovered that:
1. The ID returned by `docker image inspect --platform ...` i useless for use with other commands. (See also https://github.com/moby/moby/issues/49995) Hence, we run both a platform-specific inspect for the layer info and a platform-agnostic inspect for the image ID.
2. `ImageHistory` also has a `--platform` argument. Passing in the same platform as used for inspecting the image seems prudent.
3. `ensureLayers` calls `ImageSave` which also has a `--platform` argument. We need to pass that, as otherwise we will try to load layers for the wrong architecture.
4. `ImageLoad` and `ImageRemove` also have platform arguments. `ImageLoad` is going to autodetect the correct platform when we pass it with the tar archive; `ImageRemove` deleting the other platforms under the same tag is probably a good thing.

I tested locally with creating an `linux/amd64` builder from a lifecycle which includes the new changes.
| Exported image arch. | Old lifecycle | New lifecycle (with this PR) |
| -- | -- | -- |
| **Invoke amd64 builder with docker, amd64 host** | `linux/amd64` | `linux/amd64` |
| **Invoke amd64 builder with pack CLI, amd64 host** | `linux/amd64` | `linux/amd64` |
| **Invoke amd64 builder with docker, arm64 host** | `linux/arm64` - wrong | `linux/amd64` |
| **Invoke amd64 builder with pack CLI, arm64 host** | `linux/arm64` - wrong | `linux/amd64` |

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,6 +23,32 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+  test-and-build-linux-containerd:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Docker with Containerd
+        uses: docker/setup-docker-action@v5
+        with:
+          set-host: true
+          daemon-config: |
+            {
+              "features": {
+                "containerd-snapshotter": true
+              }
+            }
+      - name: Set up go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: Test
+        run: make test
+        env:
+          PLATFORM_AWARE_DOCKER: 'true'
+      - name: Upload Coverage to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
   test-and-build-windows:
     runs-on: windows-2022
     steps:
--- a/local/new.go
+++ b/local/new.go
@@ -3,11 +3,14 @@ package local
 import (
 	"context"
 	"fmt"
+	"runtime"
 
 	cerrdefs "github.com/containerd/errdefs"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/moby/moby/api/types/image"
 	"github.com/moby/moby/client"
+	"github.com/moby/moby/client/pkg/versions"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 
 	"github.com/buildpacks/imgutil"
 )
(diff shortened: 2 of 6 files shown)


bojidar-bg:
Dropped the ball on this for a while, but finally got around to polishing it up. Tested with a lifecycle built with the PRed version of `imgutils` on a few different platforms, and it seemed to be working. 😃 


(Unsure if the windows failure is related to my changes.)

local/new.go
```diff
@@ -64,30 +72,35 @@ func NewImage(repoName string, dockerClient DockerClient, ops ...imgutil.ImageOp
 	}, nil
 }
 
-func defaultPlatform(dockerClient DockerClient) (imgutil.Platform, error) {
+func defaultPlatform(dockerClient DockerClient) (imgutil.Platform, bool, error) {
```
jabrown85, line 75: this looks good - do you think we can get some tests with mocks? The new platform aware paths look right but tests mocking out the api version to confirm the called methods would make me feel a bit better.

.github/workflows/test.yaml
```diff
@@ -23,6 +23,28 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+  test-and-build-linux-dind:
+    runs-on: ubuntu-latest
+    services:
+      docker:
+        image: docker:dind
+        options: --privileged
+        command: dockerd --host=tcp://0.0.0.0:2376 --tls=false
```
jabrown85: this isn't valid

I think you mean

```
  services:
    docker:
      image: docker:dind
      options: --privileged
      ports:
        - 127.0.0.1:2376
```

bojidar-bg: Yeah, I had to test a bunch of things over in [my wip branch](https://github.com/bojidar-bg/imgutil/tree/l1599-local-platform-wip) before it would start cooperating 😅

local/new.go
bojidar-bg: I ended up adding a test job that uses containerd in the workflow, plus a test which only gets called in that job and should exercise the platform-aware code paths.

jabrown85 (approved):

jabrown85:
Thanks for this! I'll try and bump lifecycle to this commit and see how that lands.

bojidar-bg:
Let's hope it has better luck than last time! :sweat_smile: :sweat_smile:

PR #306, opened 2026-05-31
Add index annotation support

TIR44:
## Summary
- added `ImageIndex.SetIndexAnnotations` for top-level index annotations
- applied index annotations through go-containerregistry's index annotation mutation
- preserved top-level annotations when `SaveDir()` rewrites the local `index.json`
- updated the fake `ImageIndex`
- added unit coverage for OCI and Docker manifest list index annotations, no-annotation saves, and merge/update behavior for existing index annotation keys

## Context
This is the imgutil prerequisite for buildpacks/pack#2586. The pack-side `manifest annotate --index` flow needs an imgutil API that can write annotations to the index manifest itself, not only to a manifest descriptor inside the index.

## Tests
- `go test . -run TestCNBIndexSetIndexAnnotations -count=1`
- `go test . -run 'TestCNBIndex(SetIndexAnnotations|SaveDirWithoutIndexAnnotations)' -count=1`
- `go test . -count=1`
- `go test ./... -run '^$' -count=1`
- `make format`

GitHub checks passed on the latest commit (`6ad67aa`): DCO, Linux, Linux containerd, Windows, Codecov patch, and Codecov project.

Local Docker-dependent tests could not be run because this machine has no Docker socket at `/var/run/docker.sock`:
- `go test ./layout -run TestLayoutIndex -count=1`
- `go test ./fakes -count=1`

### DIFF
--- a/cnb_index.go
+++ b/cnb_index.go
@@ -1,6 +1,7 @@
 package imgutil
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -142,6 +143,11 @@ func (h *CNBIndex) SetAnnotations(digest name.Digest, annotations map[string]str
 	})
 }
 
+func (h *CNBIndex) SetIndexAnnotations(annotations map[string]string) (err error) {
+	h.ImageIndex = mutate.Annotations(h.ImageIndex, annotations).(v1.ImageIndex)
+	return nil
+}
+
 func (h *CNBIndex) SetArchitecture(digest name.Digest, arch string) (err error) {
 	return h.replaceDescriptor(digest, func(descriptor v1.Descriptor) (v1.Descriptor, error) {
 		descriptor.Platform.Architecture = arch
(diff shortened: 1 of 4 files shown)


cnb_index.go
```diff
@@ -272,6 +278,31 @@ func appendManifest(desc v1.Descriptor, path layout.Path, errs *SaveError) {
...
+	if len(annotations) == 0 {
+		return nil
+	}
+
+	imageIndex, err := path.ImageIndex()
+	if err != nil {
+		return err
+	}
+
+	index, err := imageIndex.IndexManifest()
```
jabrown85, line 291: Is there a reason we are re-reading `IndexManifest` when the only caller is `SaveDir` which already has pulled it?

cnb_index.go
```diff
@@ -142,6 +143,11 @@ func (h *CNBIndex) SetAnnotations(digest name.Digest, annotations map[string]str
 	})
 }
 
+func (h *CNBIndex) SetIndexAnnotations(annotations map[string]string) (err error) {
```
jabrown85, line 146: This is more of a merge than a set. Maybe change the name to match the tests?

jabrown85, line 146: nit: new methods should have doc comments

PR #308, opened 2026-07-15
Fix off-by-one bounds check in layout layer facade

arpitjain099:
The two bounds checks in `newLayerOrFacadeFrom` use `>` where they should use `>=`, so when `layerIndex` equals the slice length the guard doesn't fire and we fall straight through to `configFile.RootFS.DiffIDs[layerIndex]` / `manifestFile.Layers[layerIndex]`. That's an index out of range panic instead of the error the guard was there to return.

It's reachable through the layout facade: `newImageFacadeFrom` ranges over `original.Layers()` and hands each index to `newLayerOrFacadeFrom`, which then indexes into the config's diff_ids. Config and manifest get parsed independently and nothing cross-checks the counts, so an OCI layout image whose manifest lists more layer descriptors than the config has diff_ids (or the other way round) loads fine and then panics when it's used as a base or previous image. Only affects the facade path, so data-less/sparse layers.

Changing `>` to `>=` gets you the intended "failed to find layer for index N" error. Valid images are unaffected since a valid index never equals the array length.

Added `layout/v1_facade_test.go` covering both directions of the mismatch plus an in-bounds case. It's an in-package test since `newLayerOrFacadeFrom` isn't exported. Panics before the change, passes after, and `go test ./layout/...` is green.

### DIFF
--- a/layout/v1_facade.go
+++ b/layout/v1_facade.go
@@ -133,10 +133,10 @@ func newLayerOrFacadeFrom(configFile v1.ConfigFile, manifestFile v1.Manifest, la
 	if hasData(originalLayer) {
 		return originalLayer, nil
 	}
-	if layerIndex > len(configFile.RootFS.DiffIDs) {
+	if layerIndex >= len(configFile.RootFS.DiffIDs) {
 		return nil, fmt.Errorf("failed to find layer for index %d in config file", layerIndex)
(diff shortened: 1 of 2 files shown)


jabrown85 (approved):
Thanks for this!

PR #309, opened 2026-07-20
local: clean up daemon layer extract dirs after Save

### LINKED ISSUE #1167: Temporary files are not deleted after build
### Summary

Temporary files are not deleted after build when using with `podman`.

The `podman` doesn't support "sparse" tarballs, so `pack` has to download builder layers to temp.
The issue is that the files are not deleted after build finished.

Even more so when `pack build` is run repeatedly multiple identical temporary directories occupies temp filesystem.
This is particularly problematic when `/tmp` is in RAM.

I think `pack` should one of the following:
* delete the temp directory right after build
* do not delete the directory after build and reuse it for subsequent builds

locker95:
## Summary

When a daemon rejects sparse/omit-base-layer image loads (notably **podman**), `doDownloadLayersFor` extracts the image under `/tmp/imgutil.local.image.*` so base layers can be re-sent. Those directories were never removed, so repeated `pack build` runs left large trees in `/tmp`.

### Changes

- Track extract dirs created during download
- After `Save` / `SaveFile` consume the layers, remove those dirs (`defer` cleanup)
- On download failure, remove the extract dir immediately
- Drop only layer handles from that download; layers registered via `AddLayer` from caller-owned paths (e.g. pack builder scratch) are left alone
- Reset `downloadOnce` so a later `ensureLayers` can re-download if needed

Daemon image cache and intentional pack caches are not touched.

## Test plan

- [x] `go test ./local/ -run TestCleanupDownloadDirs`
- [ ] With podman: run `pack build` twice and confirm `/tmp/imgutil.local.image.*` does not accumulate
- [ ] Docker sparse omit-base path still works

Fixes buildpacks/pack#1167

### DIFF
--- a/local/store.go
+++ b/local/store.go
@@ -35,6 +35,12 @@ type Store struct {
 	downloadOnce         *sync.Once
 	onDiskLayersByDiffID map[v1.Hash]annotatedLayer
 	platform             imgutil.Platform
+	// downloadDirs holds temp dirs created while fetching base layers from the daemon
+	// (e.g. when sparse/omit-base-layer saves are not accepted, such as with podman).
+	// These are removed after Save/SaveFile so /tmp does not accumulate large layer extracts.
+	// Layers added explicitly via AddLayer (build artifacts) are not stored here.
+	downloadDirs      []string
+	downloadedDiffIDs []v1.Hash
 }
 
 // DockerClient is subset of client.APIClient required by this package.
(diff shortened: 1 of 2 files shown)


local/store.go
```diff
@@ -481,15 +500,45 @@ func (s *Store) doDownloadLayersFor(identifier string) error {
...
+		diffID, err := layer.DiffID()
+		if err != nil {
+			return err
+		}
+		addedDiffIDs = append(addedDiffIDs, diffID)
 	}
+
+	s.downloadDirs = append(s.downloadDirs, tmpDir)
+	s.downloadedDiffIDs = append(s.downloadedDiffIDs, addedDiffIDs...)
+	keep = true
```
jabrown85: This being set after the successful processing means that any early exit during the loop will have `onDiskLayersByDiffID` misaligned with `downloadedDiffIDs` and leave some layers dangling. Can we get a fix and a test on a not so happy path?

locker95: yeah — AddLayer was stuffing the map before we recorded the ids, so a later failure leaked handles. failing downloads now drop whatever that pass added, and there's a test for that path.

jabrown85 (approved):

